PR #35, opened 2024-08-23
added upgradeability flag and test

Ramarti:
Current contracts would require to perform an upgrade to disable upgrading.
This PR adds a simple one time flag to disable upgradeability from an admin method, so we can make the contract non upgradeable with less moving parts.

- `UpgradeabilityFlag` contract with the logic and flag
- Every UUPSUpgradeable contract inherits and uses modifier to protect `_authorizeUpgrade()`
- Unit tests for admin transfers and upgrades.


Ramarti:
Switching to Transparent Proxy instead

PR #55, opened 2024-08-26
test(evmstaking/types): add test cases for withdraw

zsystm:
This PR increased test coverage of [client/x/evmstaking/types/withdraw.go](https://github.com/piplabs/story/blob/main/client/x/evmstaking/types/withdraw.go) from 0% to 100%.

issue: #64

### DIFF
--- a/client/x/evmstaking/types/withdraw_test.go
+++ b/client/x/evmstaking/types/withdraw_test.go
@@ -8,7 +8,6 @@ import (
 
 	"github.com/cosmos/cosmos-sdk/types/module/testutil"
 	"github.com/ethereum/go-ethereum/common"
-	"github.com/stretchr/testify/require"
 	"github.com/stretchr/testify/suite"
 )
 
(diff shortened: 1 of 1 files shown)


client/x/evmstaking/types/withdraw_test.go
```diff
@@ -0,0 +1,202 @@
...
+
+func (suite *WithdrawTestSuite) TestString() {
+	require := suite.Require()
+	// Define the test cases
+	testCases := []struct {
+		name           string
+		withdrawal     types.Withdrawal
+		expectedString string
+	}{
+		{
```
limengformal, line 40: Add a test where the resulted string has leading or trailing spaces to test out this[ line](https://github.com/piplabs/story/blob/main/client/x/evmstaking/types/withdraw.go#L24C9-L24C26).

zsystm: @limengformal 
Added at https://github.com/piplabs/story/pull/55/commits/21d7ef85d84c2e4d73b6a153bc5f550889bd4d5d

client/x/evmstaking/types/withdraw_test.go
```diff
@@ -84,6 +85,11 @@ func (suite *WithdrawTestSuite) TestWithdrawalsString() {
 		suite.delAddr, suite.valAddr, suite.evmAddr.String(),
 	)
 	require.Equal(expectedString, ws.String())
+
+	// Test with spaces, it should trim the spaces
+	stringWithSpaces := " " + ws.String() + " " // add leading and trailing spaces
+	trimmedString := strings.TrimSpace(stringWithSpaces)
```
limengformal, line 91: I think we want to test that `String()` will correctly trim spaces not `TrimSpace` can trim spaces. You should add a new test case where it has leading or trailing spaces in the resulted string from a `NewWithdrawal`. If it is impossible to construct such a case, you may skip the test case.

zsystm: @limengformal 
I think it's good to skip the test case because testing that case is quite tricky. 
To test that part, we would need to mock the `String()` that intentionally returns a string with leading or trailing spaces. However, this seems quite complex and cumbersome.

limengformal: @zsystm Then let's remove it

zsystm: @limengformal 
You mean let's remove the code we are currently discussing in the comments? 
I'll remove it soon.

limengformal: yes

limengformal (approved):

PR #72, opened 2024-08-31
fix(log): remove delegation logs

edisonz0718:
remove long logs

issue: none


jdubpark (approved):

jdubpark:
I recommend we change the log level from debug to info as well for public ones.

PR #73, opened 2024-08-31
refactor(log): make abci log info

edisonz0718:
make key abci call log info so we can turn the debug level to info

issue: none


jdubpark (approved):
Why do we need these logs as info, though? I don't think we actively use these logs

edisonz0718:
> Why do we need these logs as info, though? I don't think we actively use these logs

For easier debugging. So that we know how much time each step takes. At least at the testnet it's helpful

Issue #81, opened 2024-09-05
Add test coverage of evmengine module


zsystm:
### Coverage status 

File | Cov(%) | PR | Cov after PR(%)
-- | -- | -- | --
evmengine/keeper/abci.go | 52.3 | https://github.com/piplabs/story/pull/143 | 83.8%  
evmengine/keeper/db.go | 75.0 | https://github.com/piplabs/story/pull/120 | 87.5%
evmengine/keeper/evmmsgs.go | 39.1 | https://github.com/piplabs/story/pull/95 | TBD
evmengine/keeper/genesis.go | 0.0 | #134 | 83.3%
evmengine/keeper/helpers.go | 92.3 | https://github.com/piplabs/story/pull/131 | 100%
evmengine/keeper/keeper.go | 67.2 | #132 | 89.7%
evmengine/keeper/msg_server.go | 53.1 | #100 |  86.8%
evmengine/keeper/params.go | 0.0 | https://github.com/piplabs/story/pull/136 | 77.3%
evmengine/keeper/upgrades.go | 33.3 | https://github.com/piplabs/story/pull/86 | 100%
evmengine/types/genesis.go | 0.0 | https://github.com/piplabs/story/pull/135 | 100%
evmengine/types/params.go | 0.0 | https://github.com/piplabs/story/pull/137 | 100%
evmengine/types/tx.go | 68.8 |  #83  | 100% | 
evmengine/types/upgrade_contract.go | 75.0 |  #82  | 100%

- auto generated files like `*pb.go` or `*mocks.go` are not target
- status will be updated whenever related PR or work is done

PR #95, opened 2024-09-09
test(evmengine/keeper): add test cases for evmmsgs

zsystm:
increased coverage of evmmsgs to 91.7% and do some refactoring of `evmmsgs.go` 

This PR includes #106 and will be updated based on discussion result for #106

issue: #81


zsystm:
This test code has become outdated and is no longer valid.

PR #96, opened 2024-09-09
feat(evmstaking): implement epoched staking

Narangde:
Implemention of epoched staking in `evmstaking` module.

With epoched staking, the messages are handled in 2 steps: queueing & processing. The messages from EL related to validator set are not executed immediately, but every epoch. 

### Queueing messages

When events related to validator set (`CreateValidator`, `Deposit`, `Withdraw`, `Redelegate`, and `Unjail`) are emitted from EL, the messages are queued in `k.MessageQueue`. 

### Processing messages

In `EndBlock` of `evmstaking` module, it is checked every block if the next epoch is started or not. If the next epoch is started, it executes all queued messages and return the updated validator set.

### Epoch info

The epoch is managed in `epochs` keeper, so it needs to be injected to `evmstaking` keeper. The `epoch_identifier` for epoched staking is added to params of `evmstaking`.

issue: none


client/x/epochs/module/module.proto
```diff
@@ -0,0 +1,12 @@
+syntax = "proto3";
+
+package client.x.epochs.module;
+
+import "cosmos/app/v1alpha1/module.proto";
+
+// Module is the config object of the epochs module.
+message Module {
+  option (cosmos.app.v1alpha1.module) = {
```
LeoHChen, line 9: why do you need this?

Narangde: This is basic config object for module. Other modules also have this proto file and auto-generated `module.pulsar.go`.

scripts/buf.gen.gogo.yaml
```diff
@@ -6,3 +6,5 @@ plugins:
       - plugins=grpc
       - Mgoogle/protobuf/any.proto=github.com/cosmos/cosmos-sdk/codec/types
       - Mcosmos/orm/v1/orm.proto=cosmossdk.io/orm
+      - Mcosmos/staking/v1beta1/tx.proto=github.com/cosmos/cosmos-sdk/x/staking/types
+      - Mcosmos/slashing/v1beta1/tx.proto=github.com/cosmos/cosmos-sdk/x/slashing/types
```
Narangde, line 10: This is to use `cosmos-sdk/x/staking/types` and `cosmos-sdk/x/slashing/types` instead of `cosmossdk.io/x/*` in generating pb.go file

client/x/evmengine/types/expected_keepers.go
```diff
@@ -16,8 +16,6 @@ type AccountKeeper interface {
 }
 
 type EvmStakingKeeper interface {
-	ProcessDeposit(ctx context.Context, ev *bindings.IPTokenStakingDeposit) error
-	ProcessWithdraw(ctx context.Context, ev *bindings.IPTokenStakingWithdraw) error
```
Narangde, line 20: These two are not used in `evmengine` module

Narangde:
@zsystm With this implementation, I fixed some tests of evmstaking module. Please review this [4c344d2](https://github.com/piplabs/story/pull/96/commits/4c344d20c6c545ba84f99a60288837247f75d3ce).

zsystm:
@Narangde 
Since I’m on vacation today, I won’t be able to review the updated test code. I’ll review it tomorrow.

scripts/buf.gen.gogo.yaml
edisonz0718: What is the difference between the two

client/x/evmstaking/types/params.proto
```diff
@@ -16,4 +16,8 @@ message Params {
   uint64 min_partial_withdrawal_amount = 3 [
     (gogoproto.moretags) = "yaml:\"min_partial_withdrawal_amount\""
   ];
+  // epoch_identifier mint epoch identifier e.g. (day, week).
```
edisonz0718, line 19: What does `mint` here mean?

client/x/evmstaking/types/evmstaking.proto
```diff
@@ -32,3 +35,25 @@ message Withdrawal {
...
 }
+
+// QueuedMessage is a message that can change the validator set and is delayed
+// to the end of an epoch
+message QueuedMessage {
+  // tx_id is the ID of the tx that contains the message
+  bytes tx_id = 1;
+  // msg_id is the original message ID, i.e., hash of the marshaled message
+  bytes msg_id = 2;
+  // block_height is the height when this msg is submitted to Babylon
```
edisonz0718, line 46: We should remove `Babylon` in the comments. Not just here but other places. 
Also if we reference their code, we should mention this in the README section of evmstaking module

client/x/evmstaking/keeper/abci.go
```diff
@@ -78,11 +78,6 @@ func (k *Keeper) EndBlock(ctx context.Context) (abci.ValidatorUpdates, error) {
 		}
 	}
 
-	valUpdates, err := k.stakingKeeper.EndBlocker(ctx)
```
edisonz0718, line 81: I think we cannot move valUpdates logic to the end. It has to be before adding the unbonded entries to withdraw queue. 

The reason is that: 
During staking keeper's EndBlocker, it unbonds the mature unbending delegations and release the delegated token. Only after the token release, we can do the next step, which is to burn these delegated token and prepare to mint them to EL side.

client/x/evmstaking/keeper/abci.go
```diff
@@ -132,5 +127,47 @@ func (k *Keeper) EndBlock(ctx context.Context) (abci.ValidatorUpdates, error) {
...
+
+	isNextEpoch, err := k.IsNextEpoch(ctx)
+	if err != nil {
+		return nil, errors.Wrap(err, "check if the next epoch starts")
+	}
+
+	//nolint:nestif // readability
+	if isNextEpoch {
+		// get queued messages
+		queuedMsgs, err := k.DequeueAllMsgs(ctx)
```
edisonz0718, line 141: I'm thinking to improve the performance, we can combine the DequeueAllMsgs and ProcessMsg. Because we don't actually need to dequeue and store in a separate data structure.

We just need to iterate through the queue and directly process each message. In the end, you initialize a new message queue anyway so no need to dequeue.

client/x/evmstaking/keeper/abci.go
Narangde: Ah, I see. you're right. I will move it before handling unbondedEntries.

scripts/buf.gen.gogo.yaml
Narangde: Without this option, the generated pb.go file imports the type of `staking` and `slashing` from `cosmossdk.io/x`, but no go package for the two module. So I replaced to import them from `cosmos/cosmos-sdk/x`.

client/x/evmstaking/keeper/abci.go
Narangde: Got it. valid point.
You mean, at the end of epoch, iterate all queued message with direct execution (not dequeue) and after that initialize it, right?

Narangde: Could you check this [99c399e](https://github.com/piplabs/story/pull/96/commits/99c399e230849d3e27015e4532b28faef53f48b0)?
I merge two funcs and pushed as a separate commit.

edisonz0718: Yes! This looks good to me.

client/x/evmstaking/keeper/epochs.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+	epochIdentifier := params.EpochIdentifier
+
+	// get current epoch
+	epoch, err := k.epochsKeeper.GetEpochInfo(ctx, epochIdentifier)
+	if err != nil {
+		return false, errors.Wrap(err, "get current epoch info from epochs keeper", "epoch_identifier", epochIdentifier)
+	}
+
+	// if current epoch number is greater than current epoch num, execute queued messages
```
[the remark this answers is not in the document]
edisonz0718: I think this comment needs to be rephrased

zsystm (commented):
comments for mocking usage

client/x/evmstaking/keeper/abci.go
```diff
@@ -78,9 +78,35 @@ func (k *Keeper) EndBlock(ctx context.Context) (abci.ValidatorUpdates, error) {
...
+
+	isNextEpoch, err := k.IsNextEpoch(ctx)
 	if err != nil {
-		return nil, err
+		return nil, errors.Wrap(err, "check if the next epoch starts")
+	}
+
+	if isNextEpoch {
+		// process all queued messages
+		if err := k.ProcessAllMsgs(ctx); err != nil {
```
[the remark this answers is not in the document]
jdubpark: Epoch blocks usually take longer due to backlog calculations, e.g. Osmosis's daily epoch block takes up to a few minutes at worst. We should set settings to allow epoch blocks to exceed any timeout

client/x/evmstaking/keeper/epochs.go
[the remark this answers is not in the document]
Narangde: Oops, I rephrase it.

client/x/evmstaking/keeper/abci.go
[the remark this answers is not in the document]
Narangde: Right. we need to allow longer block time for epoch block.

client/x/evmstaking/keeper/abci_test.go
```diff
@@ -370,7 +366,7 @@ func (s *TestSuite) setupMatureUnbondingDelegation(ctx sdk.Context, delAddr sdk.
 	s.setupUnbonding(pastCtx, delAddr, valAddr, amt)
 
 	// Mock staking.EndBlocker
-	s.BankKeeper.EXPECT().UndelegateCoinsFromModuleToAccount(gomock.Any(), stypes.NotBondedPoolName, delAddr, gomock.Any()).Return(nil)
+	s.BankKeeper.EXPECT().UndelegateCoinsFromModuleToAccount(gomock.Any(), stypes.NotBondedPoolName, delAddr, gomock.Any()).Return(nil).AnyTimes()
```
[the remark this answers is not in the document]
Narangde: Thank you for pointing out!
I made it much clearer by mocking call when it is needed instead of using `.AnyTimes()`.

Narangde:
I added validation of epoch duration for epoched staking in [a6c4fd5](https://github.com/piplabs/story/pull/96/commits/a6c4fd5af7c28dd746df944d9fa80be0efb5014c).

This is because if the unbonding period is shorter than the epoch duration, the unbonding is mature, but the staking keeper's `EndBlocker` may not process `completeUnbonding` and fail to send coins from the account to the module.

edisonz0718:
> I added validation of epoch duration for epoched staking in [a6c4fd5](https://github.com/piplabs/story/pull/96/commits/a6c4fd5af7c28dd746df944d9fa80be0efb5014c).
> 
> This is because if the unbonding period is shorter than the epoch duration, the unbonding is mature, but the staking keeper's `EndBlocker` may not process `completeUnbonding` and fail to send coins from the account to the module.

Is this because

client/x/evmstaking/keeper/abci.go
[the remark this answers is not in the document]
edisonz0718: How are we thinking about setting a longer timeout for epoch blocks? I think the current timeout in config is set for all blocks

client/x/evmstaking/keeper/msg_queue.go
```diff
@@ -0,0 +1,79 @@
...
+		return errors.Wrap(err, "unwrap queued msg to sdk msg")
+	}
+
+	switch unwrappedMsg := unwrappedMsgWithType.(type) {
+	case *stype.MsgCreateValidator:
+		if err := k.ProcessCreateValidatorMsg(ctx, unwrappedMsg); err != nil {
+			return errors.Wrap(err, "process CreateValidator msg")
+		}
+	case *stype.MsgDelegate:
+		if err := k.ProcessDepositMsg(ctx, unwrappedMsg); err != nil {
```
jdubpark, line 59: In the future, maybe we can add an improvement where deposits are handled per block but only go into effect at the end of each epoch.

jdubpark:
FWIW, the current design implies that withdrawing a stake doesn't immediately lower the validator power, and only starts to unbond (for 21 days) at the start of each epoch.

Narangde:
> Could you elaborate this more? Why completeUnbonding is not processed if unbonding period is shorter than the epoch duration? I thought withdrawals (full or partial) should happen regardlessly.

Because completeUnbonding is called in staking keeper's `EndBlocker`, which is called every epoch.
I will give an example of this. Let's assume the unbonding period is 10s and epoch duration is 60s. And epoch starts at time `t`.

1. `t`: epoch starts
2. `t+10` withdraw tx submitted and queued
3. `t+60` withdraw tx executed when the next epoch starts. Here, the unbonding complete time is `t+70`.
4. `t+70` unbonding is completed, but `completeUnbonding` is not called yet because epoch is not passed, leading to failure in submit coins from account to module.

client/x/evmstaking/keeper/msg_queue.go
```diff
@@ -0,0 +1,79 @@
...
+
+	for ; iterator.Valid(); iterator.Next() {
+		qMsg, err := iterator.Value()
+		if err != nil {
+			return errors.Wrap(err, "get value of message queue")
+		}
+
+		if err := k.processMsg(ctx, &qMsg); err != nil {
+			log.Warn(ctx, "Failed to process queued message", err, "tx_id", string(qMsg.TxId))
+			return errors.Wrap(err, "process queued message")
```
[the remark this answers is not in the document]
Narangde: I agree that we continue processing messages. I think it would be good to remove warn log here and return error, and just add warn log in `abci.go` and continue processing messages. Since the error returned here is used for our test codes.

Also totally agree that we need to handle failed messages. This is also the part that I had the most trouble with. Since the case where a message fails can be quite complex to consider right now, I'm thinking of storing failed messages separately rather than handling them right away, so that we can handle them later appropriately for example via upgrade. What do you think?

client/x/evmstaking/keeper/abci.go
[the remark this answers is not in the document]
Narangde: I'm not sure we can set longer timeout for epoch block, but if possible, it could be good. Why don't we make improvement in another PR?

client/x/evmstaking/keeper/keeper.go
```diff
@@ -144,7 +149,7 @@ func (k Keeper) ProcessStakingEvents(ctx context.Context, height uint64, logs []
 				continue
 			}
 			ev.StakeAmount.Div(ev.StakeAmount, gwei)
-			if err = k.ProcessCreateValidator(ctx, ev); err != nil {
+			if err = k.HandleCreateValidatorEvent(ctx, ev); err != nil {
```
edisonz0718, line 152: Can we combine all these events to a single enqueue since all of them share the same msg type

Narangde: Yes, all types of events are converted to `types.QueuedMessage` and enqueue to a single queue. Did you mean to combine all `Handle*Event` to a single func?

lib/netconf/local/genesis.json
```diff
@@ -175,8 +175,10 @@
       "params": {
         "max_withdrawal_per_block": 32,
         "max_sweep_per_block": 128,
-        "min_partial_withdrawal_amount": 100000000
-      }
+        "min_partial_withdrawal_amount": 100000000,
+        "epoch_identifier": "minute"
```
edisonz0718, line 179: How do we add epoch duration in genesis

client/x/evmstaking/keeper/withdraw.go
```diff
@@ -268,7 +270,7 @@ func (k Keeper) ProcessWithdraw(ctx context.Context, ev *bindings.IPTokenStaking
 
 	amountCoin, _ := IPTokenToBondCoin(ev.Amount)
 
-	log.Debug(ctx, "Processing EVM staking withdraw",
+	log.Info(ctx, "EVM staking withdraw detected",
```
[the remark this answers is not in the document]
Narangde: For other events, the log level in `Info` when the event is detected, thus I changed this to `Info` as well for code consistency. I think `Debug` is enough when the event is detected. How about we changing all log for event detection to `Debug` level?

client/x/evmstaking/keeper/abci.go
```diff
@@ -28,108 +28,134 @@ func (k *Keeper) EndBlock(ctx context.Context) (abci.ValidatorUpdates, error) {
...
 
-		// push the undelegation to the withdrawal queue
-		err = k.AddWithdrawalToQueue(ctx, types.NewWithdrawal(
-			uint64(blockHeight),
-			entry.delegatorAddress,
-			entry.validatorAddress,
-			delEvmAddr,
-			entry.amount.Uint64(),
-		))
+		partialWithdrawals, err := k.ExpectedPartialWithdrawals(ctx)
```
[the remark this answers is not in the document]
Narangde: You're right.

lib/netconf/local/genesis.json
Narangde: Only epoch identifier is needed in evmstaking module. Epoch info such as epoch duration is in `epochs` module, and it can be added in genesis.json for now. There is no way to add epoch info as cosmos-sdk do.

client/x/evmstaking/keeper/keeper.go
Narangde: I don't think we can combine all `Handle*Event` function to a single function. Because each event has different fields. If we want to make a single function, we need to move the logics for handling events (converting each event to `QueuedMessage` and enqueuing) out to `ProcessStakingEvents`. If I misunderstand your comment, please correct me.

lib/netconf/local/genesis.json
```diff
@@ -175,8 +175,10 @@
...
+      },
+      "epoch_number": "0"
```
limengformal, line 181: Is this the starting number of epoch?

lib/netconf/local/genesis.json
limengformal: It will be great to add configuration in genesis.json to change the number of blocks per epoch. It helps to perform tests in reasonable amount of time.

0xHansLee: The epoch duration can be set in `epochs` module. For local or mininet test, we can use shorter epoch by setting shorter epoch. `emvstaking` module depends on `epochs` module

lib/netconf/local/genesis.json
0xHansLee: It is the epoch number currently used in `evmstaking` module for epoch staking. This value is used for checking the next epoch is started in `epochs` module by comparing epoch number.

lib/netconf/local/genesis.json
edisonz0718: If we can set the duration in genesis.json, it will make setting different networks easier. Otherwise, different networks need to manage different binaries.

0xHansLee: I think it's ok if each network has different genesis.json file (to be exact, different epoch identifier for epoch staking).

If we add duration in `evmstaking` param (and add to genesis.json), we don't need `epochs` module in `evmstaking` module and we don't need `epochs` module in Story for now. `evmstaking` module can be independent with `epochs` module. The reason why I have added and used `epochs` module for epoch staking is that I wanna make `epochs` module manage epoch stuffs in our app for modularity. Do you think that it better than current design that `evmstaking` module have independent epoch info for epoch staking?

client/x/evmstaking/keeper/abci.go
```diff
@@ -28,99 +28,125 @@ func (k *Keeper) EndBlock(ctx context.Context) (abci.ValidatorUpdates, error) {
...
+			return nil, errors.Wrap(err, "update validator set")
 		}
 
-		// This should not produce error, as all delegations are done via the evmstaking module via EL.
-		// However, we should gracefully handle in case Get fails.
-		delEvmAddr, err := k.DelegatorMap.Get(ctx, entry.delegatorAddress)
-		if err != nil {
-			return nil, errors.Wrap(err, "map delegator pubkey to evm address")
+		// init message queue
+		if err := k.MessageQueue.Initialize(ctx); err != nil {
```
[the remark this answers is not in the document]
0xHansLee: Yes right.

lib/netconf/local/genesis.json
edisonz0718: Yeah I think we should add epoch info we use for evmstaking to epochs module's genesis file. I guess what you meant is that it should be in another PR?

edisonz0718 (approved):

client/x/evmstaking/keeper/withdraw.go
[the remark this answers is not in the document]
0xHansLee: @edisonz0718, could you check this comment? As you said, it seems like to be better that we use `Debug` for all tx log.

lib/netconf/local/genesis.json
0xHansLee: Done in https://github.com/piplabs/story/pull/181

client/x/evmstaking/keeper/abci.go
```diff
@@ -28,99 +28,125 @@ func (k *Keeper) EndBlock(ctx context.Context) (abci.ValidatorUpdates, error) {
...
+				if entry.IsMature(ctxTime) && !entry.OnHold() {
+					// track undelegation only when remaining or truncated shares are non-zero
+					if !entry.Balance.IsZero() {
+						amt := sdk.NewCoin(bondDenom, entry.Balance)
+						// TODO: check if it's possible to add a double entry in the unbondedEntries array
+						unbondedEntries = append(unbondedEntries, UnbondedEntry{
+							validatorAddress: dvPair.ValidatorAddress,
+							delegatorAddress: dvPair.DelegatorAddress,
+							amount:           amt.Amount,
+						})
```
limengformal, line 83: I thought we have a limit on the number of operations per unbonding period or epoch. Do we know where that is enforced?

client/x/evmstaking/keeper/abci.go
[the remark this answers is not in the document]
limengformal: We should definitely increase the timeout. I would imagine a lot of operations need to be processed during the epoch block. Can be in another PR

client/x/evmstaking/keeper/abci.go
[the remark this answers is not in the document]
limengformal: Do we initialize every epoch? I thought there is a max amount that one epoch can process and the rest will flow to the next epoch

limengformal (approved):

client/x/evmstaking/keeper/abci.go
0xHansLee: As commented [here](https://github.com/piplabs/story/pull/96#discussion_r1794884092), no limit in this PR. Will do in the following PR.

0xHansLee:
> Based on the comments, there are still two changes to be made:
>
> We need a maximum length for the message queue. Can set it to 40,000 for now. Also, we only process X requests from the queue per epoch. We can set it to 4096 first.
Increase the timeout for the epoch block.
You can create issues and add them in a separate PR if it's a bigger change.

Thank you for your clarification. I will make another PR for the above things.

PR #100, opened 2024-09-10
test(evmengine/keeper): add test cases for msg server

zsystm:
increased coverage of `evmengine/keeper/msg_server.go` to 86.8% without considering un-used code `compareWithdrawals`.

### Changes
**`mockEngineAPI`**
- `forceInvalidNewPayloadV3` and `forceInvalidForkchoiceUpdatedV3` are added to simulate failed engine apis with invalid status return.

**`engineMock`**
- `storeKey` is added to make engineMock dependent on `sdk.Context` for better testability.
- By using `storeKey`, engineMock's methods can interact with the `sdk.Context`'s store, allowing for independent tests that do not interfere with each other’s store state. 
  - now we can use cacheContext to test engineMock

**`NewBlock`**
- Because of above changes, we need to rlp encode and decode block data. But if we create a block with nil withdrawals and nil withdrawalHash, rlp doesn't work well (couldn't figure out the root cause)
- To avoid rlp encoding error, pass non-nil withdrawals so withdrawalHash can be set as non-nil emptyHash value.

### Not covered
- empty app hash case
- sdk related
  - failed to update execution head which uses sdk's ormtable
- geth edge cases
  - engine api returns response with un-known PayloadStatus

issue: #81

### DIFF
--- a/client/x/evmengine/keeper/abci_internal_test.go
+++ b/client/x/evmengine/keeper/abci_internal_test.go
@@ -124,7 +124,7 @@ func TestKeeper_PrepareProposal(t *testing.T) {
 		for _, tt := range tests {
 			t.Run(tt.name, func(t *testing.T) {
 				t.Parallel()
-				ctx, storeService := setupCtxStore(t, nil)
+				ctx, storeKey, storeService := setupCtxStore(t, nil)
 				cdc := getCodec(t)
 				txConfig := authtx.NewTxConfig(cdc, nil)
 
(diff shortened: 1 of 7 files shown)


edisonz0718:
It's a bit weird to add store to engineMock since it should represent only the EngineAPI. Can you point me to the part of the tests that requires us to do this?

zsystm:
@edisonz0718 

In Test_msgServer_ExecutionPayload, during the pass case, engineMock.ForkchoiceUpdatedV3 is called. (ref. [code](https://github.com/zsystm/story/blob/zsystm/test/evmengine/keeper/msg_server/client/x/evmengine/keeper/msg_server_internal_test.go#L251-L255))

The original implementation of engineMock simply updates its head after ForkchoiceUpdatedV3 is executed. This means we can’t revert the state to test other scenarios. Without adding context, we have no way to mock the Execution Layer’s rollback.

I considered creating a new `engineMock` for each test case, but introducing context (ctx) seemed more flexible and convenient for managing state changes. Using context allows us to isolate test cases more easily, making it simpler to write and manage future tests without needing to repeatedly create new mocks.

edisonz0718 (approved):

limengformal (approved):

Issue #64, opened 2024-08-27
Add test coverage of evmstaking module


zsystm:
### Coverage status 

| File                                  | Cov(%) | PR  | Cov after PR(%) |
|---------------------------------------|--------|-----|-----|
| evmstaking/keeper/abci.go             | 0.0    | https://github.com/piplabs/story/pull/79 | 84.5% |
| evmstaking/keeper/deposit.go          | 0.0    | https://github.com/piplabs/story/pull/71 | 94.4% |
| evmstaking/keeper/genesis.go          | 0.0    | https://github.com/piplabs/story/pull/111 | 71.4% |
| evmstaking/keeper/grpc_query.go       | 0.0    | #164 | 78.9% |
| evmstaking/keeper/keeper.go           | 11.4   | https://github.com/piplabs/story/pull/75 | 94.3% |
| evmstaking/keeper/msg_server.go       | 0.0    | https://github.com/piplabs/story/pull/70 | 95.5% | 
| evmstaking/keeper/params.go           | 0.0    | https://github.com/piplabs/story/pull/102 | 70% |
| evmstaking/keeper/redelegation.go     | 72.4   | https://github.com/piplabs/story/pull/77 | 100% |
| evmstaking/keeper/set_address.go      | 0.0    | https://github.com/piplabs/story/pull/67 | 87.5% |
| evmstaking/keeper/staking_queue.go    | 0.0    | https://github.com/piplabs/story/pull/68 | 84.6% |
| evmstaking/keeper/unjail.go           | 0.0    | https://github.com/piplabs/story/pull/104 | 100% | 
| evmstaking/keeper/utils.go            | 100.0  | https://github.com/piplabs/story/pull/76 |  100% | 
| evmstaking/keeper/validator.go        | 0.0    | https://github.com/piplabs/story/pull/66 | 86.0% |
| evmstaking/keeper/withdraw.go         | 0.0    | https://github.com/piplabs/story/pull/65 | 86.8% |
| evmstaking/keeper/withdrawal_queue.go | 0.0    | https://github.com/piplabs/story/pull/61 | 78.8% |
| evmstaking/types/genesis.go           | 0.0    |  https://github.com/piplabs/story/pull/112 | 100% |
| evmstaking/types/params.go            | 0.0    | https://github.com/piplabs/story/pull/103 | 100% |
| evmstaking/types/staking_contract.go  | 75.0   | https://github.com/piplabs/story/pull/56 |  100% |
[...]

PR #103, opened 2024-09-11
test(evmstaking/types): add test cases for params

zsystm:
increased test coverage to 100% 

**changes**
- Updated the input types of validation functions to use concrete types instead of any, as there doesn’t appear to be a specific reason for using any. (Please let me know if there’s a reason for it)
- Renamed the withdrawal test suite to avoid naming conflicts with the parameter test suite.

issue: #64


jdubpark (approved):

edisonz0718 (approved):
lgtm

PR #106, opened 2024-09-11
chore(evmengine): refactor evmmsgs

### LINKED ISSUE #105: Refactor evmEvents for Better Code Quality
## Description and context
The current implementation of [evmEvents](https://github.com/piplabs/story/blob/main/client/x/evmengine/keeper/evmmsgs.go#L17-L71) contains multiple logical operations directly within the function, which should be separated into distinct functions. 
- Fetching logs from ethereum client
- Converting those logs to EVMEvent objects
- Verifying the converted events
- Sorting the events

This approach limits testability, as it prevents writing tests for specific logic handled by `evmEvents` and increases complexity by combining too many responsibilities.

## Suggested solution
Refactor by delegating separable logic into dedicated functions. For example, we can serpate the following logics to other functions.
- Geth log conversion and verification
- Sorting EVMEvents

Current: the `evmEvents` function handled
- Fetching logs from ethereum client
- Converting those logs to `EVMEvent` objects
- Verifying the converted events
- Sorting the events

Suggestion: The `evmEvents` function 
- Fetch logs from ethereum client
- Delegates conversion and verification to `types.EthLogToEVMEvent`.
- Delegates sorting to `types.SortEVMEvents`.

## Definition of done
#106 is merged

zsystm:
refactored evmmsgs for better code quailty and testability.

issue: #105

### DIFF
--- a/client/x/evmengine/types/tx.go
+++ b/client/x/evmengine/types/tx.go
@@ -87,37 +87,3 @@ func SortEVMEvents(events []*EVMEvent) {
 		return bytes.Compare(events[i].Data, events[j].Data) < 0
 	})
 }
-
-// IsSortedEVMEvents check if the events are sorted by ascending order of address, topics, and data.
-func IsSortedEVMEvents(events []*EVMEvent) bool {
-	for i := 1; i < len(events); i++ {
-		// Compare addresses first
-		addressComparison := bytes.Compare(events[i-1].Address, events[i].Address)
-		if addressComparison > 0 {
-			// it is not sorted by ascending order of address
-			return false
-		}
-
-		if addressComparison == 0 {
-			// If addresses are equal, compare by topics
-			previousTopic := slices.Concat(events[i-1].Topics...)
-			currentTopic := slices.Concat(events[i].Topics...)
-			topicComparison := bytes.Compare(previousTopic, currentTopic)
-
-			if topicComparison > 0 {
-				// it is not sorted by ascending order of topics
-				return false
-			}
-
-			if topicComparison == 0 {
-				// If topics are also equal, compare by data
-				if bytes.Compare(events[i-1].Data, events[i].Data) > 0 {
-					// it is not sorted by ascending order of data
-					return false
-				}
-			}
-		}
-	}
-
-	return true
-}
--- a/client/x/evmengine/types/tx.go
+++ b/client/x/evmengine/types/tx.go
@@ -1,6 +1,10 @@
 package types
 
 import (
+	"bytes"
+	"slices"
+	"sort"
+
 	"github.com/ethereum/go-ethereum/common"
 	ethtypes "github.com/ethereum/go-ethereum/core/types"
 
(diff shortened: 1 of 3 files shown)


client/x/evmengine/types/tx_test.go
```diff
@@ -162,3 +169,270 @@ func TestEVMEvent_Verify(t *testing.T) {
...
+
+func TestSortEVMEvents(t *testing.T) {
+	t.Parallel()
+	stakingAbi, err := bindings.IPTokenStakingMetaData.GetAbi()
+	require.NoError(t, err, "failed to load ABI")
+	stakingAddr := common.HexToAddress(predeploys.IPTokenStaking)
+	slashingAbi, err := bindings.IPTokenSlashingMetaData.GetAbi()
+	require.NoError(t, err, "failed to load ABI")
+	slashingAddr := common.HexToAddress(predeploys.IPTokenSlashing)
+	require.Negative(t, bytes.Compare(stakingAddr.Bytes(), slashingAddr.Bytes()), "stakingAddr should be less than slashingAddr")
```
jdubpark, line 270: What do you mean `stakingAddr should be less than slashingAddr` for a compare? (+ same for below error messages)

jdubpark (approved):

client/x/evmengine/types/tx_test.go
zsystm: Thanks for the feedback! The reason for checking that `stakingAddr` is less than `slashingAddr` is to ensure a known initial order before testing the sorting logic. 
This precondition is necessary because we want to verify that if the addresses are input in the wrong order, the sorting function will correctly reorder them. 
Similarly, the same check is applied for the event topics (withdrawEv and depositEv) to confirm the function’s behavior.

To make this clearer, I’ve added more detailed comments in the code explaining why these comparisons are necessary and included examples to illustrate the expected behavior. (commit: https://github.com/piplabs/story/pull/106/commits/5027d58e7053f92f5ae329cd74dc8a73e575e295) 
Please take a look at the updated comments, and let me know if there’s anything else I can clarify.

client/x/evmengine/types/tx.go
```diff
@@ -47,3 +51,73 @@ func (l *EVMEvent) Verify() error {
...
+		if cmp := bytes.Compare(topicI, topicJ); cmp != 0 {
+			return cmp < 0
+		}
+
+		return bytes.Compare(events[i].Data, events[j].Data) < 0
+	})
+}
+
+// IsSortedEVMEvents check if the events are sorted by ascending order of address, topics, and data.
+func IsSortedEVMEvents(events []*EVMEvent) bool {
```
edisonz0718, line 92: Is this only used in the test? If so, we should move it to the test

zsystm: @edisonz0718 

applied it in https://github.com/piplabs/story/pull/106/commits/828cb405aa5786f3324a346a24d55b965cb1cf03

thanks for your comment.

jdubpark (approved):

client/x/evmengine/types/tx.go
zsystm: @edisonz0718 
Could you review this PR when you have time? I applied your comment. Thanks in advance.

zsystm:
This patch code has become outdated and is no longer valid.

PR #125, opened 2024-09-17
chore(evmstaking): remove unnecessary codes

Narangde:
Remove unnecessary codes of getting all delegations in `EndBlock` of `evmstaking`.
I think this was for debugging of `evmstaking` module.

issue: none


jdubpark (approved):
good catch

PR #126, opened 2024-09-18
feat(x/signal): signal module for single binary upgrade

jdubpark:
single binary upgrade using CIP-10

issue: none


client/app/blocked.go
LeoHChen, line 1: don't combine features together please. OFAC option shall be in a separate PR.

PR #129, opened 2024-09-18
feat(docker): add docker node support

LuttyYang:
According to the #153 requirements, we need to support starting the RPC node with Docker. This PR implements the functionality to start a Story's RPC node and localnet using Docker.

issue: #153


docker/localnet/script/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM ubuntu:22.04
```
LeoHChen, line 1: can you use a lightweight base image?

LuttyYang: Coinbase requires using Ubuntu 20 or higher as the base image, rather than Alpine or Debian.

jdubpark:
Is there a way for us to unify modifications to the genesis and config files in `node-launcher` with the genesis and config files in 1+1, 1+2, and 1+4? A lot of moving pieces otherwise.

edisonz0718:
The localnet is set up in another repo

PR #142, opened 2024-09-23
feat(script): auto add binary version to s3 file

jhd2best:
this feature will automatic add version record when new version built. which can let developer reset or upgrade network nodes more convenient

issue: none


LuttyYang (approved):
LGTM

PR #154, opened 2024-09-29
feat(x/evmstaking): max unbond withdrawal

jdubpark:
max(delegator balance, unbond request amount)

issue: none

### DIFF
--- a/client/x/evmstaking/keeper/abci.go
+++ b/client/x/evmstaking/keeper/abci.go
@@ -95,8 +95,9 @@ func (k *Keeper) EndBlock(ctx context.Context) (abci.ValidatorUpdates, error) {
 			return nil, errors.Wrap(err, "delegator address from bech32")
 		}
 
-		maxAmount := k.bankKeeper.GetBalance(ctx, delegatorAddr, sdk.DefaultBondDenom).Amount
-		if entry.amount.LT(maxAmount) {
+		maxAmount := k.bankKeeper.SpendableCoin(ctx, delegatorAddr, sdk.DefaultBondDenom).Amount
+		maxBounded := entry.amount.LT(maxAmount)
+		if maxBounded {
 			maxAmount = entry.amount
 		}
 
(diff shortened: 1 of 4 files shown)


client/x/evmstaking/keeper/abci.go
```diff
@@ -110,6 +111,16 @@ func (k *Keeper) EndBlock(ctx context.Context) (abci.ValidatorUpdates, error) {
 			return nil, errors.Wrap(err, "burn coins")
 		}
 
+		log.Debug(ctx, "Adding undelegation to withdrawal queue",
+			"delegator", entry.delegatorAddress,
+			"validator", entry.validatorAddress,
+			"max_amount", maxAmount.String())
+		if entry.amount.LT(maxAmount) {
+			log.Debug(ctx, "Undelegation amount is less than max amount",
```
LeoHChen, line 119: This log can be a warning then.
And log the full information of the delegator and validator here

jdubpark: 58fc5ed

client/x/evmstaking/keeper/abci.go
```diff
@@ -110,6 +111,19 @@ func (k *Keeper) EndBlock(ctx context.Context) (abci.ValidatorUpdates, error) {
 			return nil, errors.Wrap(err, "burn coins")
 		}
 
+		log.Debug(ctx, "Adding undelegation to withdrawal queue",
+			"delegator", entry.delegatorAddress,
+			"validator", entry.validatorAddress,
+			"max_amount", maxAmount.String())
+		if entry.amount.LT(maxAmount) {
+			log.Warn(ctx, "Undelegation amount is less than max amount",
```
kingster-will, line 119: Seems the check can never return True and the log would never be printed out, because the `maxAmount` was set to the value of `entry.amount` at Ln 99 - 101, hence the `entry.amount` can never less than `maxAmount`. sounds right?

It appears that the check will never return True, and the log will never be printed because `maxAmount` is assigned the value of `entry.amount` on lines 99–101. As a result, `entry.amount` can never be less than `maxAmount`. Does that sound correct?

jdubpark: Thanks, fixed in dfdd687

LeoHChen (approved):

Narangde:
I think we need to check if the mature unbonding is deleted or not before processing `unbondedEntries`.
If `CompleteUnbonding` fails in the staking keeper's EndBlock for any reason, the mature unbonding will not be deleted and it will be queried again in the next block's `GetMatureUnbondedDelegations`. Then, it will burn an amount of tokens of mature unbonding even if a spendable amount of token has already been burnt.

edisonz0718:
Close this PR temporarily to avoid confusion

PR #157, opened 2024-10-01
feat(cmd): one block rollback 

### LINKED ISSUE #144: Recover geth and story clients when geth client is upgraded after the upgrade height
## Description and context
In the event of a geth hard fork upgrade, if a node is not upgraded after the upgrade block height, the story client may panic since it may not reach consensus with the rest of the nodes that already have their geth clients upgraded.

Only upgrading the geth client at this point doesn't help since the story client already verified/proposed a block that is in conflict. Node at this point can only remove data folder and sync from genesis block which takes a long time.

## Suggested solution
One suggested solution is to roll back the block with incorrect proposal so the node can sync the correct block.

## Definition of done
Node can roll back incorrect block and restart with correct block in the event of a late geth upgrade.

zsystm:
add one block rollback cmd (ref. [cosmos-sdk](https://github.com/cosmos/cosmos-sdk/blob/6a0358607411cd83d22c6f1deea69f399f803007/server/rollback.go#L13-L14))

issue: #144

### DIFF
--- a/client/app/start.go
+++ b/client/app/start.go
@@ -184,6 +184,49 @@ func Start(ctx context.Context, cfg Config) (func(context.Context) error, error)
 	}, nil
 }
 
+func CreateApp(ctx context.Context, cfg Config) *App {
+	privVal, err := loadPrivVal(cfg)
+	if err != nil {
+		panic(errors.Wrap(err, "load validator key"))
+	}
+
+	db, err := dbm.NewDB("application", cfg.BackendType(), cfg.DataDir())
+	if err != nil {
+		panic(errors.Wrap(err, "create db"))
+	}
+
+	baseAppOpts, err := makeBaseAppOpts(cfg)
+	if err != nil {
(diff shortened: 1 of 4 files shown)


client/app/start.go
```diff
@@ -184,6 +184,49 @@ func Start(ctx context.Context, cfg Config) (func(context.Context) error, error)
 	}, nil
 }
 
+func CreateApp(ctx context.Context, cfg Config) *App {
```
edisonz0718, line 187: I think this function is pretty much the same code as the Start function above. Can we add a TODO to make sure the Start function uses this function later. To make it consistent

edisonz0718 (approved):

jdubpark (approved):

client/app/start.go
zsystm: @edisonz0718 
You are right. I added To-Do in https://github.com/piplabs/story/pull/157/commits/fdc427d8c6ab021cc10f2df1d6b56882e3821207

PR #171, opened 2024-10-05
refactor(panic): remove unused or unnecessary panic code

edisonz0718:
Clean up some unused code. Change unnecessary panic in Wrap function.

issue: none


client/x/evmstaking/types/withdraw.go
```diff
@@ -48,28 +44,3 @@ func NewWithdrawalFromMsg(msg *MsgAddWithdrawal) Withdrawal {
...
-func MustUnmarshalWithdrawal(cdc codec.BinaryCodec, value []byte) Withdrawal {
-	withdrawal, err := UnmarshalWithdrawal(cdc, value)
-	if err != nil {
-		panic(err)
-	}
-
-	return withdrawal
-}
-
-// UnmarshalWithdrawal returns the withdrawal.
```
jdubpark, line 67: We don't use this? Interesting, wonder why we have it in the first place

jdubpark (approved):

client/x/evmstaking/types/withdraw.go
edisonz0718: Yeah, I did not find any usage in the code base.

PR #178, opened 2024-10-09
feat(cli): support odyssey testnet

LuttyYang:
support `--odyssey` flag for automatic generate config and key

issue: none


lib/evmchain/evmchain.go
```diff
@@ -17,6 +17,9 @@ const (
 	// Testnets.
 	IDIliad uint64 = 1513
 
+	// Odyssey Testnets.
+	IDOdyssey uint64 = 1514
```
0xHansLee, line 21: Are we gonna use the same chain ID with mainnet for Odyssey?
I think it would better to use diff chain ID for testnet.

lib/netconf/odyssey/genesis.json
```diff
@@ -0,0 +1,636 @@
...
+      "outstanding_rewards": [],
+      "validator_accumulated_commissions": [],
+      "validator_historical_rewards": [],
+      "validator_current_rewards": [],
+      "delegator_starting_infos": [],
+      "validator_slash_events": []
+    },
+    "evmengine": {
+      "params": {
+        "execution_block_hash": "9evG8Jgui/f9Uys/lZ+E0Ss97dKCevjTH1OJRHvtr8Y="
```
0xHansLee, line 203: It would be helpful to add linked PR in story-geth.
maybe https://github.com/piplabs/story-geth/pull/29?

lib/evmchain/evmchain.go
leeren: We are reserving 1514 for mainnet

What about 1515?

lib/netconf/odyssey/genesis.json
leeren: ++

lib/evmchain/evmchain.go
0xHansLee: 1516 looks good

LuttyYang: changed to 1516

jdubpark (approved):

edisonz0718 (approved):

PR #181, opened 2024-10-09
feat(epochs): add default epoch info to genesis config

0xHansLee:
Added several default epoch info by default into genesis.json. The added epochs are as below:
- minute
- hour
- day
- week

The default start time is set as `1970-01-01T00:00:00Z`, so when you start app locally the epoch would increase every block. If you want to test epoch if it is proceeded every duration, you need to modify `start_time` near the time now.

issue: none

### DIFF
--- a/lib/netconf/local/genesis.json
+++ b/lib/netconf/local/genesis.json
@@ -194,7 +194,34 @@
           "current_epoch": "0",
           "current_epoch_start_time": "1970-01-01T00:00:00Z",
           "epoch_counting_started": false,
-          "current_epoch_start_height": "10"
+          "current_epoch_start_height": "1"
+        },
+        {
+          "identifier": "hour",
+          "start_time": "1970-01-01T00:00:00Z",
+          "duration": "1h",
+          "current_epoch": "0",
+          "current_epoch_start_time": "1970-01-01T00:00:00Z",
+          "epoch_counting_started": false,
+          "current_epoch_start_height": "1"
+        },
+        {
+          "identifier": "day",
+          "start_time": "1970-01-01T00:00:00Z",
+          "duration": "24h",
+          "current_epoch": "0",
+          "current_epoch_start_time": "1970-01-01T00:00:00Z",
+          "epoch_counting_started": false,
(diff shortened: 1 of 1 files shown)


lib/netconf/local/genesis.json
```diff
@@ -194,7 +194,34 @@
...
+          "current_epoch_start_height": "1"
+        },
+        {
+          "identifier": "week",
+          "start_time": "1970-01-01T00:00:00Z",
+          "duration": "168h",
+          "current_epoch": "0",
+          "current_epoch_start_time": "1970-01-01T00:00:00Z",
+          "epoch_counting_started": false,
+          "current_epoch_start_height": "1"
```
edisonz0718, line 224: Is current epoch-related information calculated based on the start time and duration of the epoch?

lib/netconf/local/genesis.json
```diff
@@ -194,7 +194,34 @@
...
+        },
+        {
+          "identifier": "week",
+          "start_time": "1970-01-01T00:00:00Z",
+          "duration": "168h",
+          "current_epoch": "0",
+          "current_epoch_start_time": "1970-01-01T00:00:00Z",
+          "epoch_counting_started": false,
```
edisonz0718, line 223: Do people need to set this to true if they plan to enable this epoch info?

0xHansLee: It indicates if the epoch timer begin. If it false, `epochs` module initialize the epoch (set current epoch 1 and begin timer). People can start epoch timer with `epoch_counting_started` of false

lib/netconf/local/genesis.json
0xHansLee: Right. This is that [part](https://github.com/piplabs/story/blob/36a50a2ac21c044dc981c0ce5ef51176929d6975/client/x/epochs/keeper/abci.go#L31-L31)

0xHansLee:
@edisonz0718  cc. @ezreal1997 
Close this PR since we decided to revert epoch module.

PR #182, opened 2024-10-09
feat(mint): update mint param type

gezerui0124:
fix mint param type issue

issue: none


client/x/mint/keeper/genesis_test.go
```diff
@@ -62,7 +63,7 @@ func (s *GenesisTestSuite) TestImportExportGenesis() {
 	genesisState := types.DefaultGenesisState()
 	genesisState.Params = types.NewParams(
 		"testDenom",
-		24625000_000_000_000,
```
edisonz0718, line 65: Should we add decimal for testing too

limengformal (approved):

edisonz0718 (approved):

PR #185, opened 2024-10-09
feat(proto): refactor proto and scripts

jdubpark:
refactor proto to best practice

issue: none


Makefile
```diff
@@ -32,7 +31,7 @@ lint: ## Runs linters via pre-commit.
 	@pre-commit run -v --all-files
 
 .PHONY: bufgen
-bufgen: ## Generates protobufs using buf generate.
```
edisonz0718, line 35: What is the reason to move to proto-gen

scripts/protocgen.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
```
edisonz0718, line 1: Does this replace buf_generate.sh? It seems to only process gogo files

Issue #184, opened 2024-10-09
Modifications of staking contracts


Ramarti:
## Description and context
- Moving `unjail()` to `IPTokenStaking` and removing `IPTokenSlashing` contract
- Remove support for redelegation
- Remove all state that would require sync with CL, and that is not configuration
- Support locked tokens.
- Support 4 different staking periods
-- 3 are incrementally longer periods, which will have a unique `delegationId`
-- 1 is flexible (no unstake restrictions, unless locked). `delegationId=0`

## Suggested solution
- Incremental counter for `delegationId`

## Definition of done
Contracts pass test locally and in devnets.

PR #189, opened 2024-10-11
feat(staking): implement new staking design changes

Ramarti:
- removes slashing contract, unjail goes to staking
- staking periods
- delegation id
- token type

issue: #184


contracts/script/GenerateAlloc.s.sol
```diff
@@ -159,37 +159,33 @@ contract GenerateAlloc is Script {
         vm.resetNonce(tmp);
 
         InitializableHelper.disableInitializers(impl);
-        IPTokenStaking(Predeploys.Staking).initialize(protocolAdmin, 1 ether, 1 ether, 1 ether, 7 days);
+        IIPTokenStaking.InitializerArgs memory args = IIPTokenStaking.InitializerArgs({
+            owner: protocolAdmin,
+            minStakeAmount: 1 ether,
```
limengformal, line 164: 1024 ether

contracts/script/GenerateAlloc.s.sol
```diff
@@ -159,37 +159,33 @@ contract GenerateAlloc is Script {
...
+            minUnstakeAmount: 1 ether,
```
limengformal, line 165: 1024 ether

contracts/script/GenerateAlloc.s.sol
```diff
@@ -159,37 +159,33 @@ contract GenerateAlloc is Script {
...
+            minCommissionRate: 5_00, // 5% in basis points
+            shortStakingPeriod: 1 days, // TBD
+            mediumStakingPeriod: 2 days, // TBD
+            longStakingPeriod: 3 days, // TBD
```
limengformal, line 169: Can this be updated in the staking contract after initialization?

contracts/script/GenerateAlloc.s.sol
```diff
@@ -159,37 +159,33 @@ contract GenerateAlloc is Script {
...
+            unjailFee: 1 ether
+        });
+
+        // Testnet timing values
+        if (block.chainid != MAINNET_CHAIN_ID) {
+            args.minCommissionRate = 5_00 seconds;
```
limengformal, line 175: it should be 5%?

contracts/script/GenerateAlloc.s.sol
```diff
@@ -159,37 +159,33 @@ contract GenerateAlloc is Script {
...
+            args.shortStakingPeriod = 10 seconds;
+            args.mediumStakingPeriod = 15 seconds;
+            args.longStakingPeriod = 20 seconds;
```
limengformal, line 178: where are these periods be used?

contracts/script/GenerateAlloc.s.sol
Ramarti: In which chain Id? testnets too?

contracts/script/GenerateAlloc.s.sol
Ramarti: In which chain Id? testnets too?

contracts/script/GenerateAlloc.s.sol
Ramarti: Yes, we can:
```
function setStakingPeriods(uint32 short, uint32 medium, uint32 long) external onlyOwner
```

contracts/script/GenerateAlloc.s.sol
Ramarti: They are passed to the initializer, and used to resolve staking duration in `stake`

limengformal (approved):

contracts/src/protocol/IPTokenStaking.sol
```diff
@@ -320,231 +327,274 @@ contract IPTokenStaking is IIPTokenStaking, Ownable2StepUpgradeable, ReentrancyG
...
     function stake(
         bytes calldata delegatorUncmpPubkey,
-        bytes calldata validatorCmpPubkey
+        bytes calldata validatorUncmpPubkey,
+        IIPTokenStaking.StakingPeriod stakingPeriod,
+        bytes calldata data
     )
         external
         payable
+        override
```
kingster-will, line 390: The override modifier might not be necessary here, since this contract doesn’t inherit from any other contract.

contracts/src/protocol/IPTokenStaking.sol
```diff
@@ -320,231 +327,274 @@ contract IPTokenStaking is IIPTokenStaking, Ownable2StepUpgradeable, ReentrancyG
...
-
-        _redelegate(delegatorCmpPubkey, p.validatorCmpSrcPubkey, p.validatorCmpDstPubkey, stakeAmount);
+        if (keccak256(validatorUncmpSrcPubkey) == keccak256(validatorUncmpDstPubkey)) {
+            revert Errors.IPTokenStaking__RedelegatingToSameValidator();
+        }
+        (uint256 stakeAmount, ) = roundedStakeAmount(msg.value);
+        if (stakeAmount < minStakeAmount) {
+            revert Errors.IPTokenStaking__StakeAmountUnderMin();
+        }
+        emit Redelegate(delegatorUncmpPubkey, validatorUncmpSrcPubkey, validatorUncmpDstPubkey, delegationId, amount);
```
kingster-will, line 477: It would be add comments to explain why we don't need to refund remainder of roundedStakeAmount().

kingster-will: If I understand correctly, the remainder will stay in the source validator. If that’s the case, we might explain how do we handle the remainder in the source validator, especially since it’s likely that the remainder will be less than the `minStakeAmount`. we might better add comments to explain why it is expected state/behavior.

PR #202, opened 2024-10-15
feat(singularity): add singularity to mint and evmstaking module

0xHansLee:
Added singularity to mint and evmstaking module.
Until singularity height, no reward is minted and no validator set is updated.

BTW, there is one thing to discuss. What if user make requests for unstake greater than `MaxEntries`? In CL side, when there are more unstake requests than `MaxEntries`, it is processed as processing failure. However, in EL, unstake is processed normally, so I think we need to handle this. What do you think? I would like to discuss this because as the Singularity period gets longer, this possibility will increase. 

issue: none


client/x/mint/types/params.go
```diff
@@ -26,7 +27,8 @@ func DefaultParams() Params {
 	return Params{
 		MintDenom:         sdk.DefaultBondDenom,
 		InflationsPerYear: math.LegacyNewDec(24625000000000000.000000000000000000),
-		BlocksPerYear:     uint64(60 * 60 * 8766 / 5), // assuming 5 second block times
+		BlocksPerYear:     uint64(60 * 60 * 8766 / 5), // assuming 5 seconds block times
+		SingularityHeight: 725760,                     // 42 days with assuming 5 seconds block time
```
edisonz0718, line 31: Should we combine this with the other SingularityHeight in evmstaking to make it consistent?

0xHansLee: If possible, it would be better to combine into one, but I have no idea on combining this. Or one way for this is that we make a new keeper named `singularity` and import the keeper to both `mint` and `evmstaking` keeper. I think there is no need to make a new keeper for this, thus make like this. Do you have any better idea on this?

client/x/evmstaking/keeper/redelegation.go
```diff
@@ -15,6 +15,16 @@ import (
 )
 
 func (k Keeper) ProcessRedelegate(ctx context.Context, ev *bindings.IPTokenStakingRedelegate) error {
+	isInSingularity, err := k.IsSingularity(ctx)
+	if err != nil {
+		return errors.Wrap(err, "check if it is singularity")
+	}
+
+	if isInSingularity {
+		log.Info(ctx, "Relegation event detected, but it is not processed since current block is singularity")
```
edisonz0718, line 24: Should we change it to debug?

client/x/evmstaking/keeper/withdraw.go
```diff
@@ -244,6 +244,16 @@ func (k Keeper) EnqueueEligiblePartialWithdrawal(ctx context.Context, withdrawal
 }
 
 func (k Keeper) ProcessWithdraw(ctx context.Context, ev *bindings.IPTokenStakingWithdraw) error {
+	isInSingularity, err := k.IsSingularity(ctx)
+	if err != nil {
+		return errors.Wrap(err, "check if it is singularity")
+	}
+
+	if isInSingularity {
+		log.Info(ctx, "Withdraw event detected, but it is not processed since current block is singularity")
```
edisonz0718, line 253: change to debug?

edisonz0718 (approved):

client/x/evmstaking/keeper/redelegation.go
0xHansLee: Yes, changed in [2150314](https://github.com/piplabs/story/pull/202/commits/2150314811ef720fd1cf5ed4bb9e5ce7e5614da3)

client/x/evmstaking/keeper/withdraw.go
0xHansLee: changed in [2150314](https://github.com/piplabs/story/pull/202/commits/2150314811ef720fd1cf5ed4bb9e5ce7e5614da3)

Issue #197, opened 2024-10-14
Replace staking module's endblocker with a function that return mature unbonded entries


edisonz0718:
## Description and context
Currently, the staking module's endblocker doesn't return the mature unbonded entries that are unbonded during the call. It makes our processing of mature unbonded in the evmstaking module messy. It also causes our most recent outage on Iliad testnet, where the actually unbondings are different from the unbondings that we think will happen. We need a solution to return accurate mature unbonding.

## Suggested solution
Add a new function in staking keeper that does all the staking endblocker does and at the same time return the mature unbonding entries that actually get unbonded.

## Definition of done
The change is deployed and tested

PR #209, opened 2024-10-16
feat(evmstaking): replace staking keeper endblocker

0xHansLee:
With [this implementation](https://github.com/piplabs/cosmos-sdk/pull/4), I replaced the EndBlocker of staking keeper. 
It now returns the array of completed unbonding, so we can use them for processing withdrawal.

issue: #197

### DIFF
--- a/client/x/evmstaking/keeper/abci.go
+++ b/client/x/evmstaking/keeper/abci.go
@@ -4,8 +4,6 @@ import (
 	"context"
 	"time"
 
-	"cosmossdk.io/math"
-
 	abci "github.com/cometbft/cometbft/abci/types"
 	"github.com/cosmos/cosmos-sdk/telemetry"
 	sdk "github.com/cosmos/cosmos-sdk/types"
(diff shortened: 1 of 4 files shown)


jdubpark (approved):

client/x/evmstaking/keeper/abci.go
```diff
@@ -29,96 +21,28 @@ func (k *Keeper) EndBlock(ctx context.Context) (abci.ValidatorUpdates, error) {
...
-		// the total spendable amount.
-		if entry.amount.GT(spendableAmount) {
-			entry.amount = spendableAmount
-			log.Warn(ctx, "Spendable amount is less than the requested undelegation amount",
-				errors.New("spendable amount is less than the requested undelegation amount"),
-				"delegator", entry.delegatorAddress,
-				"validator", entry.validatorAddress,
-				"requested_amount", entry.amount.String(),
-				"spendable_amount", spendableAmount.String())
-		}
```
jdubpark, line 113: Should we add tests to ensure this invariance is not broken? cc @limengformal 

We should also add the crisis module for invariance checks, cc @edisonz0718.

edisonz0718 (approved):
lgtm

client/x/evmstaking/keeper/abci.go
0xHansLee: Will add a test to cover that this invariance is not broken in the following PR.

PR #210, opened 2024-10-16
chore(staking): add new staking genesis config

0xHansLee:
Added a genesis configuration for local network according to https://github.com/piplabs/cosmos-sdk/pull/2.

issue: none


jdubpark (approved):

0xHansLee:
Close this PR since we decided not to set as genesis config, but hardcode it.

PR #224, opened 2024-10-17
fix(staking): removing period values from staking contract

Ramarti:
Removes period durations from smart contract so it's not duplicated in EL ad CL.

issue: none

### DIFF
--- a/contracts/src/protocol/IPTokenStaking.sol
+++ b/contracts/src/protocol/IPTokenStaking.sol
@@ -125,7 +125,6 @@ contract IPTokenStaking is IIPTokenStaking, Ownable2StepUpgradeable, ReentrancyG
         _setMinUnstakeAmount(newMinUnstakeAmount);
     }
 
-
     /// @notice Sets the fee charged for adding to CL storage.
     /// @param newFee The new fee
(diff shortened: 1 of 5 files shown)


contracts/src/protocol/IPTokenStaking.sol
```diff
@@ -455,13 +426,19 @@ contract IPTokenStaking is IIPTokenStaking, Ownable2StepUpgradeable, ReentrancyG
         if (stakeAmount < minStakeAmount) {
             revert Errors.IPTokenStaking__StakeAmountUnderMin();
         }
-        uint32 duration = 0;
         uint256 delegationId = 0;
         if (stakingPeriod != IIPTokenStaking.StakingPeriod.FLEXIBLE) {
```
limengformal, line 430: Can we verify that stakingPeriod has values that are accepted by the `enum StakingPeriod` ?

Ramarti: Good point

Ramarti: Fixed 71d5c45

limengformal (approved):

PR #226, opened 2024-10-18
feat(genesis): set genesis validator not using genutils

0xHansLee:
Set state of genesis validator in genesis.json file for local devnet, not using genutils.
The address, `story1fl48vsnmsdzcv85q5d2q4z5ajdha8yu3h6nymd`, is the address of bonded pool. This balance should be equal to the bonded coins.

issue: none


0xHansLee:
As we decided to use genutils for setting genesis validators, I close this PR.

PR #228, opened 2024-10-18
feat(staking): ubi pool contract

Ramarti:
Adds a predeploy to set the % of validator UBI in CL, collect the ubi fees, and distribute monthly

issue: none


contracts/src/protocol/IPTokenStaking.sol
```diff
@@ -50,31 +51,8 @@ contract IPTokenStaking is IIPTokenStaking, Ownable2StepUpgradeable, ReentrancyG
...
-        if (uncmpPubkey[0] != 0x04) {
-            revert Errors.IPTokenStaking__InvalidPubkeyPrefix();
-        }
-        if (_uncmpPubkeyToAddress(uncmpPubkey) != expectedAddress) {
-            revert Errors.IPTokenStaking__InvalidPubkeyDerivedAddress();
-        }
-        _;
-    }
+    /// @notice Staking periods and their corresponding durations
+    mapping(IIPTokenStaking.StakingPeriod period => uint32 duration) public stakingDurations;
```
0xHansLee, line 55: This looks like an unintentional repair.

contracts/src/interfaces/IUBIPool.sol
```diff
@@ -0,0 +1,39 @@
...
+    /// @notice Emitted when the UBI percentage is set
+    /// @param percentage The percentage of the UBI
+    event UBIPercentageSet(uint32 percentage);
+
+    /// @notice Emitted when the UBI distribution is set
+    /// @param month The month of the distribution (counter, starting from 1, not timestamp or month of year)
+    /// @param totalUBI The total amount of UBI
+    /// @param validatorUncmpPubKeys The validator uncompressed public keys
+    /// @param percentages The percentages of the UBI for each validator
+    event UBIDistributionSet(uint256 month, uint256 totalUBI, bytes[] validatorUncmpPubKeys, uint256[] percentages);
```
0xHansLee, line 16:
```suggestion
    /// @param amounts The amounts of UBI for each validator
    event UBIDistributionSet(uint256 month, uint256 totalUBI, bytes[] validatorUncmpPubKeys, uint256[] amounts);
```

contracts/src/protocol/UBIPool.sol
```diff
@@ -0,0 +1,81 @@
...
+    /// @param month The month of the distribution (counter, starting from 1, not timestamp or month of year)
+    /// @param totalUBI The total amount of UBI
+    /// @param validatorUncmpPubKeys The validator uncompressed public keys
+    /// @param amounts The amounts of UBI for each validator
+    function setUBIDistribution(
+        uint256 month,
+        uint256 totalUBI,
+        bytes[] calldata validatorUncmpPubKeys,
+        uint256[] calldata amounts
+    ) external onlyOwner {
```
0xHansLee, line 52: Don't we need to validate each uncompressed pubkey is valid?

contracts/src/protocol/IPTokenStaking.sol
limengformal, line 55: Is this used?

contracts/src/protocol/UBIPool.sol
Ramarti: Good point

contracts/src/interfaces/IUBIPool.sol
jdubpark: Need to clarify the units, is it relative (%) or absolute (token units).

contracts/src/protocol/UBIPool.sol
```diff
@@ -0,0 +1,81 @@
...
+/// @title UBIPool
+/// @notice Contract for distributing UBI to validators
+/// UBI comes from a percentage of the protocol's emmission during a defined period
+/// This % is minted to the UBIPool contract and can be claimed by validators
+/// Each validator can claim their UBI for a given month
+/// Distributions will be made public monthly by Story for community scrutiny,
+/// and if correct, will be distributed to validators via this contract
+contract UBIPool is IUBIPool, Ownable2StepUpgradeable, ReentrancyGuardUpgradeable, PubKeyVerification {
+    /// @notice The maximum UBI percentage
+    uint32 public immutable MAX_UBI_PERCENTAGE;
```
jdubpark, line 18: What's the decimal exponent for this (since Solidity doesn't support decimals)? 100% => 100_0000? Need to ensure we have enough accuracy in UBI %

contracts/src/protocol/UBIPool.sol
jdubpark: I agree, but to save gas, we could do it off-chain.

contracts/src/protocol/UBIPool.sol
```diff
@@ -0,0 +1,81 @@
...
+    function initialize(address owner) public initializer {
+        require(owner != address(0), "UBIPool: owner cannot be zero address");
+        __Ownable_init(owner);
+    }
+
+    /// @notice Sets the UBI percentage distribution in CL
+    /// @param percentage The percentage of the UBI
+    function setUBIPercentage(uint32 percentage) external onlyOwner {
+        require(percentage < MAX_UBI_PERCENTAGE, "UBIPool: percentage too high");
+        emit UBIPercentageSet(percentage);
```
jdubpark, line 39: This doesn't set any storage variable, only emits the event.

contracts/src/protocol/UBIPool.sol
```diff
@@ -0,0 +1,81 @@
...
+    /// @notice Sets the UBI distribution for a given month
+    /// @param month The month of the distribution (counter, starting from 1, not timestamp or month of year)
+    /// @param totalUBI The total amount of UBI
+    /// @param validatorUncmpPubKeys The validator uncompressed public keys
+    /// @param amounts The amounts of UBI for each validator
+    function setUBIDistribution(
+        uint256 month,
+        uint256 totalUBI,
+        bytes[] calldata validatorUncmpPubKeys,
+        uint256[] calldata amounts
```
kingster-will, line 51: Since the function can override the amount for a month if it has already been set, it may be better to add a flag, such as `bool override`, to explicitly allow overriding the amount if desired. Alternatively, we could simply revert the transaction to prevent any accidental overrides.

contracts/src/interfaces/IUBIPool.sol
kingster-will, line 16: it would be better to rename `percentages` to `amounts` to be consistent with implementation contract.

contracts/src/protocol/UBIPool.sol
0xHansLee: Is there any limitations on `setUBIPercentage` even if this function can be called only by owner?

contracts/src/interfaces/IUBIPool.sol
```diff
@@ -0,0 +1,39 @@
...
+    /// @param month The month of the distribution (counter, starting from 1, not timestamp or month of year)
+    /// @param totalUBI The total amount of UBI
+    /// @param validatorUncmpPubKeys The validator uncompressed public keys
+    /// @param percentages The percentages of the UBI for each validator
+    function setUBIDistribution(
+        uint256 month,
+        uint256 totalUBI,
+        bytes[] calldata validatorUncmpPubKeys,
+        uint256[] calldata percentages
+    ) external;
```
kingster-will, line 32: Same as above, rename `percentages` to `amounts`.

kingster-will (commented):
Looks good. just couple comments.

contracts/src/protocol/UBIPool.sol
Ramarti: fixed

contracts/src/protocol/IPTokenStaking.sol
Ramarti: Good catch, this got there by rebasing I guess, previous PRs removed it

Ramarti: Good catch, this got there by rebasing I guess, previous PRs removed it

contracts/src/protocol/UBIPool.sol
Ramarti: I was using 2 decimals. What's the preferred amount by CL @ezreal1997 ?

contracts/src/protocol/UBIPool.sol
Ramarti: @jdubpark 
> This doesn't set any storage variable, only emits the event.
Yes, CL module reads the event and acts on it

@0xHansLee 
> Is there any limitation on setUBIPercentage even if this function can be called only by owner?
On the number of times is called? No, it wasn't defined. Better not to overcomplicated, in case governance needs to fix an amount, it must be set to 0 for some emergency, etc

contracts/src/protocol/UBIPool.sol
Ramarti: We decided to use an internal counter in the contract as distribution id instead of `month`, less chances for input error.
Editing complicates things, there must be careful review of this tx before setting it (maybe timelock).

contracts/src/interfaces/IUBIPool.sol
Ramarti: Done, thanks for noticing!

contracts/src/protocol/UBIPool.sol
Ramarti: 2 decimals is fine for this

edisonz0718 (approved):

0xHansLee (approved):

PR #245, opened 2024-10-19
feat(genesis): update odyssey genesis file

edisonz0718:
Update Odyssey genesis file

issue: none


0xHansLee (commented):
- Why don't we change the `genesis_time` in L1?
- Why don't we set initial balance to 1M gwei?
- Should the commission rate of each validator be greater than `min_commission_rate` in staking module's param?
- Why don't we add `support_token_type` to each messages for creating validator?

0xHansLee (approved):

PR #249, opened 2024-10-19
feat(evmstaking): using pubkey instead address

jhd2best:
- all staking and distribution API interface use pubkey instead of address
- add get delegator info API interface

issue: none


0xHansLee (approved):
LGTM, excepting the failure of go lint action

edisonz0718 (approved):

PR #260, opened 2024-10-21
feat(evmstaking): separate withdrawal and reward queue

ezreal1997:
separate withdrawal and reward queue, which withdrawal queue has higher priority

issue: none

### DIFF
--- a/client/x/evmengine/keeper/abci.go
+++ b/client/x/evmengine/keeper/abci.go
@@ -57,11 +57,11 @@ func (k *Keeper) PrepareProposal(ctx sdk.Context, req *abci.RequestPreparePropos
 	if err != nil {
 		return nil, errors.Wrap(err, "get max withdrawal per block")
 	}
-	withdrawals, err := k.evmstakingKeeper.PeekEligibleWithdrawals(ctx, int(maxWithdrawals))
+	withdrawals, err := k.evmstakingKeeper.PeekEligibleWithdrawals(ctx, maxWithdrawals)
 	if err != nil {
 		return nil, errors.Wrap(err, "error on withdrawals dequeue")
 	}
-	maxRewardWithdrawals := int(maxWithdrawals) - len(withdrawals)
+	maxRewardWithdrawals := maxWithdrawals - uint32(len(withdrawals))
 	rewardWithdrawals, err := k.evmstakingKeeper.PeekEligibleRewardWithdrawals(ctx, maxRewardWithdrawals)
 	if err != nil {
 		return nil, errors.Wrap(err, "error on reward withdrawals dequeue")
@@ -206,12 +206,12 @@ func (k *Keeper) PostFinalize(ctx sdk.Context) error {
 		log.Error(ctx, "Starting optimistic build failed; get max withdrawal", err, logAttr)
 		return errors.Wrap(err, "get max withdrawal per block")
 	}
-	withdrawals, err := k.evmstakingKeeper.PeekEligibleWithdrawals(ctx, int(maxWithdrawals)) // context is "context.TODO()"" (empty) in CometBFT v0.38
+	withdrawals, err := k.evmstakingKeeper.PeekEligibleWithdrawals(ctx, maxWithdrawals) // context is "context.TODO()"" (empty) in CometBFT v0.38
 	if err != nil {
 		log.Error(ctx, "Starting optimistic build failed; withdrawals peek", err, logAttr)
 		return nil
 	}
-	maxRewardWithdrawals := int(maxWithdrawals) - len(withdrawals)
+	maxRewardWithdrawals := maxWithdrawals - uint32(len(withdrawals))
 	rewardWithdrawals, err := k.evmstakingKeeper.PeekEligibleRewardWithdrawals(ctx, maxRewardWithdrawals)
 	if err != nil {
 		log.Error(ctx, "Starting optimistic build failed; reward withdrawals peek", err, logAttr)
--- a/client/x/evmengine/keeper/abci.go
+++ b/client/x/evmengine/keeper/abci.go
@@ -53,10 +53,20 @@ func (k *Keeper) PrepareProposal(ctx sdk.Context, req *abci.RequestPreparePropos
 
 	appHash := common.BytesToHash(ctx.BlockHeader().AppHash)
 
-	withdrawals, err := k.evmstakingKeeper.PeekEligibleWithdrawals(ctx)
+	maxWithdrawals, err := k.evmstakingKeeper.MaxWithdrawalPerBlock(ctx)
+	if err != nil {
+		return nil, errors.Wrap(err, "get max withdrawal per block")
+	}
+	withdrawals, err := k.evmstakingKeeper.PeekEligibleWithdrawals(ctx, int(maxWithdrawals))
 	if err != nil {
 		return nil, errors.Wrap(err, "error on withdrawals dequeue")
 	}
+	maxRewardWithdrawals := int(maxWithdrawals) - len(withdrawals)
+	rewardWithdrawals, err := k.evmstakingKeeper.PeekEligibleRewardWithdrawals(ctx, maxRewardWithdrawals)
+	if err != nil {
+		return nil, errors.Wrap(err, "error on reward withdrawals dequeue")
+	}
+	withdrawals = append(withdrawals, rewardWithdrawals...)
 
 	// Either use the optimistic payload or create a new one.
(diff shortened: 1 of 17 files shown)


client/x/evmengine/keeper/proposal_server.go
```diff
@@ -90,30 +91,67 @@ func evmEventsEqual(a, b []*types.EVMEvent) error {
...
-		if withdrawal.Address != expectedWithdrawals[i].Address {
+		if expectedRewardWithdrawals[i].Address != actualWithdrawals[pos].Address {
 			return errors.New("invalid withdrawal address")
 		}
-		if withdrawal.Amount != expectedWithdrawals[i].Amount {
+		if expectedRewardWithdrawals[i].Amount != actualWithdrawals[pos].Amount {
 			return errors.New("invalid withdrawal amount")
 		}
+		pos++
 	}
```
jdubpark, line 155: Can compare concurrently

client/x/evmstaking/keeper/withdrawal_queue.go
```diff
@@ -55,10 +53,9 @@ func (k Keeper) DequeueEligibleWithdrawals(ctx context.Context) (withdrawals ety
 	return withdrawals, nil
 }
 
-func (k Keeper) PeekEligibleWithdrawals(ctx context.Context) (withdrawals etypes.Withdrawals, err error) {
-	maxDequeue, err := k.MaxWithdrawalPerBlock(ctx)
-	if err != nil {
-		return nil, err
+func (k Keeper) PeekEligibleWithdrawals(ctx context.Context, maxPeek int) (withdrawals etypes.Withdrawals, err error) {
+	if maxPeek <= 0 {
```
jdubpark, line 57: No need to check for negative? Or do we expect `maxPeek` to be a negative value at some point? This should probably be an invariance check (crisis module) for each block.

client/x/evmengine/keeper/abci.go
```diff
@@ -53,10 +53,20 @@ func (k *Keeper) PrepareProposal(ctx sdk.Context, req *abci.RequestPreparePropos
...
+	maxWithdrawals, err := k.evmstakingKeeper.MaxWithdrawalPerBlock(ctx)
+	if err != nil {
+		return nil, errors.Wrap(err, "get max withdrawal per block")
+	}
+	withdrawals, err := k.evmstakingKeeper.PeekEligibleWithdrawals(ctx, int(maxWithdrawals))
 	if err != nil {
 		return nil, errors.Wrap(err, "error on withdrawals dequeue")
 	}
+	maxRewardWithdrawals := int(maxWithdrawals) - len(withdrawals)
+	rewardWithdrawals, err := k.evmstakingKeeper.PeekEligibleRewardWithdrawals(ctx, maxRewardWithdrawals)
```
jdubpark, line 65: If we move to an epoch-based withdrawal system later, it could be possible to flood the stake withdrawal queue s.t. `maxRewardWithdrawals = 0` for many epochs, which delays reward receiving and could affect reward assumptions, e.g. restaking protocols. Thus, an option to consider is whether making `maxRewardWithdrawals` dependent on `maxWithdrawals` (ie. maxStakeWithdrawals) or not. cc @0xHansLee @edisonz0718

jdubpark (approved):

client/x/evmengine/types/expected_keepers.go
```diff
@@ -19,11 +19,14 @@ type AccountKeeper interface {
 type EvmStakingKeeper interface {
 	ProcessDeposit(ctx context.Context, ev *bindings.IPTokenStakingDeposit) error
 	ProcessWithdraw(ctx context.Context, ev *bindings.IPTokenStakingWithdraw) error
```
0xHansLee, line 21: Could you delete these two as well? As I remember, these two are not used in evmengine keeper.

client/x/evmstaking/keeper/withdrawal_queue.go
```diff
@@ -17,12 +17,10 @@ func (k Keeper) AddWithdrawalToQueue(ctx context.Context, withdrawal types.Withd
 	return k.WithdrawalQueue.Enqueue(ctx, withdrawal)
 }
 
-func (k Keeper) DequeueEligibleWithdrawals(ctx context.Context) (withdrawals etypes.Withdrawals, err error) {
-	maxDequeue, err := k.MaxWithdrawalPerBlock(ctx)
-	if err != nil {
-		return nil, err
+func (k Keeper) DequeueEligibleWithdrawals(ctx context.Context, maxDequeue int) (withdrawals etypes.Withdrawals, err error) {
```
0xHansLee, line 20: We can simply change like this and no need to check if the value is negative. maxDequeue is a param of max withdrawal per block of evmengine module. 

Also, in the below for loop in this function, no need to wrap `maxDequeue` with `uint64` I think.
```suggestion
func (k Keeper) DequeueEligibleWithdrawals(ctx context.Context, maxDequeue uint32) (withdrawals etypes.Withdrawals, err error) {
```

client/x/evmengine/keeper/abci.go
0xHansLee: That's true. We may need other approach to withdraw these when epoch system is introduced.

client/x/evmstaking/keeper/withdrawal_queue.go
0xHansLee, line 20: Why don't we add break condition when looping the queue? If the queue is empty, just break the loop.

client/x/evmengine/keeper/proposal_server.go
```diff
@@ -90,30 +91,67 @@ func evmEventsEqual(a, b []*types.EVMEvent) error {
...
+		)
+	}
+
+	maxRewardWithdrawals := int(maxWithdrawals) - len(expectedWithdrawals)
+	expectedRewardWithdrawals, err := s.evmstakingKeeper.PeekEligibleRewardWithdrawals(ctx, maxRewardWithdrawals)
+	if err != nil {
+		return errors.Wrap(err, "peek reward withdrawals")
+	}
+
+	localNum := len(expectedWithdrawals) + len(expectedRewardWithdrawals)
```
edisonz0718, line 117: the name localNum is not very clear to me

client/x/evmstaking/keeper/withdraw.go
```diff
@@ -308,9 +307,6 @@ func (k Keeper) EnqueueEligibleRewardWithdrawal(ctx context.Context, rewardWithd
 		}
 	}
 
-	// set metrics
-	promutil.EVMStakingQueueDepth.Set(float64(k.WithdrawalQueue.Len(ctx)))
```
edisonz0718, line 312: Why not replace this with the reward queue metrics?

client/x/evmengine/keeper/msg_server.go
```diff
@@ -45,12 +45,55 @@ func (s msgServer) ExecutionPayload(ctx context.Context, msg *types.MsgExecution
...
 		return nil, errors.Wrap(err, "error on withdrawals dequeue")
 	}
-	log.Debug(ctx, "Dequeueing eligible withdrawals [AFTER]", "len", len(payload.Withdrawals))
+	log.Debug(
+		ctx, "Dequeueing eligible withdrawals [AFTER]",
+		"total_len", len(payload.Withdrawals),
+		"withdrawals_len", len(ws),
+	)
+
+	if len(ws) > len(payload.Withdrawals) {
```
edisonz0718, line 67: Why this should be >

client/x/evmengine/keeper/msg_server.go
```diff
@@ -45,12 +45,55 @@ func (s msgServer) ExecutionPayload(ctx context.Context, msg *types.MsgExecution
...
 	// TODO: We might need to check that the withdrawals in the finalized block are the same as the ones dequeued.
-	// But there's no way to reject the block at this point, so we can only log an error.
-	log.Debug(ctx, "Dequeueing eligible withdrawals [BEFORE]", "len", len(payload.Withdrawals))
-	if _, err := s.evmstakingKeeper.DequeueEligibleWithdrawals(ctx); err != nil {
+	// Since we already checked the withdrawals in the proposal server, we simply check the length here.
+	log.Debug(
+		ctx, "Dequeueing eligible withdrawals [BEFORE]",
+		"total_len", len(payload.Withdrawals),
+	)
+	maxWithdrawals, err := s.evmstakingKeeper.MaxWithdrawalPerBlock(ctx)
```
edisonz0718, line 53: I think we should add queue validation part to a separate function

edisonz0718 (approved):

PR #265, opened 2024-10-21
feat(evidence): add evidence module with singularity

0xHansLee:
With [this](https://github.com/piplabs/cosmos-sdk/pull/18), slashing for double signing is also disabled during singularity.

issue: none

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -296,10 +296,10 @@ replace (
 	cosmossdk.io/core v0.12.0 => cosmossdk.io/core v0.11.0
 
 	// replace evidence module to the story's evidence module
-	cosmossdk.io/x/evidence => github.com/piplabs/cosmos-sdk/x/evidence v0.1.2-0.20241021201424-b11be3a942ed
+	cosmossdk.io/x/evidence => github.com/piplabs/cosmos-sdk/x/evidence v0.1.2-0.20241021233425-b586999d3a6a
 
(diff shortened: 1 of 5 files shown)


jdubpark (approved):

go.mod
```diff
@@ -294,8 +295,11 @@ replace (
 	// should use v0.11.0. The Cosmos build fails with types/context.go:65:29: undefined: comet.BlockInfo otherwise.
 	cosmossdk.io/core v0.12.0 => cosmossdk.io/core v0.11.0
 
+	// replace evidence module to the story's evidence module
+	cosmossdk.io/x/evidence => github.com/piplabs/cosmos-sdk/x/evidence v0.1.2-0.20241021201424-b11be3a942ed
+
 	// Direct cosmos-sdk branch link: https://github.com/piplabs/cosmos-sdk/tree/piplabs/v0.50.7, current branch: piplabs/v0.50.7
-	github.com/cosmos/cosmos-sdk => github.com/piplabs/cosmos-sdk v0.50.7-piplabs-v0.14
+	github.com/cosmos/cosmos-sdk => github.com/piplabs/cosmos-sdk v0.50.8-0.20241021201424-b11be3a942ed
```
jdubpark, line 302: Ultimately, we should use direct import instead of replacing

0xHansLee: Even if we use direct import for cosmos-sdk, we may need to this kind of replacement I think. Because other packages, which imports cosmos-sdk (e.g. our modified evidence module), also import cosmos-sdk in the code base.

edisonz0718 (approved):

PR #276, opened 2024-10-22
feat(genesis): predeploy erc6551 registry

kingster-will:
update genesis

issue: none

### DIFF
--- a/contracts/script/GenerateAlloc.s.sol
+++ b/contracts/script/GenerateAlloc.s.sol
@@ -176,7 +176,7 @@ contract GenerateAlloc is Script {
         // predeploys that are not upgradable
         setCreate3();
         deployTimelock();
-        deployERC6551();
+        setERC6551();
 
         // predeploys that are upgradable
         setProxy(Predeploys.Staking);
@@ -334,7 +334,7 @@ contract GenerateAlloc is Script {
         console2.log("Create3 deployed at:", Predeploys.Create3);
     }
 
-    function deployERC6551() internal {
+    function setERC6551() internal {
         address tmp = address(new ERC6551Registry());
         vm.etch(Predeploys.ERC6551Registry, tmp.code);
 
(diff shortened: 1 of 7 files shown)


Ramarti (commented):
LGTM, but please add a test file with at least a simple test to see if `ERC6551Registry` is there and working at genesis time.

Please check /contracts/test/utils/Test.sol to understand the setup, thanks!

limengformal (approved):
As @Ramarti mentioned we want to add tests

kingster-will:
Added more tests.

contracts/script/GenerateAlloc.s.sol
```diff
@@ -332,6 +334,19 @@ contract GenerateAlloc is Script {
         console2.log("Create3 deployed at:", Predeploys.Create3);
     }
 
+    function setERC6551() internal {
+        address tmp = address(new ERC6551Registry());
```
Ramarti, line 338: Have we verified that `ERC6551Registry` does not set storage values in its constructor?
If it does, we have to set the same values in the correct addresses, using `vm.store`

contracts/src/libraries/Predeploys.sol
```diff
@@ -19,6 +19,9 @@ library Predeploys {
     /// @dev We maximize compatibility with the contracts deployed by ZeframLou
     address internal constant Create3 = 0x9fBB3DF7C40Da2e5A0dE984fFE2CCB7C47cd0ABf;
 
+    /// @notice ERC6551Registry address
```
Ramarti, line 22: Please add the reason we are not using a namespace address (0xCCCC...), for example a link to the ERC.

contracts/script/GenerateAlloc.s.sol
kingster-will: Confirmed.  ERC6551Registry has no any storage.

contracts/src/libraries/Predeploys.sol
kingster-will: Added comments.

Ramarti (approved):
LGTM!

PR #278, opened 2024-10-23
feat(genesis): reserve address space for 1024 proxies

Ramarti:
Proxies will be initialized to a deterministic but dead address. They can be upgraded later, or renounce upgradeability.
ProxyAdmins will be owned by the timelock.

issue: none

### DIFF
--- a/contracts/script/GenerateAlloc.s.sol
+++ b/contracts/script/GenerateAlloc.s.sol
@@ -178,14 +178,20 @@ contract GenerateAlloc is Script {
         deployTimelock();
         setERC6551();
 
-        // predeploys that are upgradable
-        setProxy(Predeploys.Staking);
-        setProxy(Predeploys.Upgrades);
-        setProxy(Predeploys.UBIPool);
+        // Set proxies for all predeploys
+        setProxies();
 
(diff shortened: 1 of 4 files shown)


contracts/script/GenerateAlloc.s.sol
```diff
@@ -178,16 +178,22 @@ contract GenerateAlloc is Script {
...
         setUBIPool();
     }
 
+    function setProxies() internal {
+        for (uint160 i = 1; i <= Predeploys.NamespaceSize; i++) {
+            address addr = address(uint160(Predeploys.Namespace) + i);
+            setProxy(addr);
+        }
+    }
+
```
kingster-will, line 196: Better add comments to describe the purpose of the function that preserve proxies address for future use. for better readability.

0xHansLee (approved):

contracts/script/GenerateAlloc.s.sol
Ramarti: Added

kingster-will (approved):
LGTM!

limengformal (approved):

PR #279, opened 2024-10-23
fix(evmstaking): fixed get delegator address

jhd2best:
fixed get delegator address

issue: none


client/server/staking.go
```diff
@@ -294,27 +294,29 @@ func (s *Server) GetDelegatorByDelegatorAddress(r *http.Request) (resp any, err
...
+		"delegator_addr":   delAddr,
+		"withdraw_address": "",
+		"reward_address":   "",
+		"operator_address": "",
+	}
+
 	delWithdrawEvmAddr, err := s.store.GetEvmStakingKeeper().DelegatorWithdrawAddress.Get(queryContext, delAddr)
-	if err != nil {
-		return nil, err
+	if err == nil {
```
edisonz0718, line 305: I think we should still handle errors here.

edisonz0718: Only handle specific errors we want to handle differently

client/server/staking.go
```diff
@@ -294,27 +294,29 @@ func (s *Server) GetDelegatorByDelegatorAddress(r *http.Request) (resp any, err
 
 	delAddr := mux.Vars(r)["delegator_addr"]
 
+	result := map[string]string{
```
edisonz0718, line 297: We should use a struct with a proper naming here. By default, string type is "".

client/server/struct.go
```diff
@@ -0,0 +1,8 @@
+package server
```
edisonz0718, line 1: Can we name this file to type.go? Just to be consistent

edisonz0718 (approved):

PR #282, opened 2024-10-23
feat(wip): introduce wrapped ip and predeploy

kingster-will:
Introduce Wrapped IP and predeploy WIP in genesis

issue: none


0xHansLee (approved):
lgtm

limengformal (approved):

Ramarti (approved):
LGTM, but I believe we should re introduce checking transfer to zero address.

Keeping the discussion here, we can do it in a different PR
https://github.com/piplabs/story/issues/283

PR #285, opened 2024-10-23
chore(wip): prevent transfer to zero or wip

### LINKED ISSUE #283: decision: should WIP prevent transfer to zero address?
- Our ERC20 uses Solady, a heavily optimized ERC20 implementation compared to OpenZeppelin's more opinionated implementation.
- Part of this optimization is removing UX/security checks in multiple operations (self approval, minting to zero address, etc). 
- One of the removals is the check that allows transferring tokens to `address(0)`

Do we want to add the latter check back?
- Story will be a consumer chain, with less technical users
- A bug in a wallet or frontend that forgets to initialize `to` input parameter in `transfer` could mean that users lose their assets, transferring to zero by a mistake they have no agency over.
- Transfer to zero doesn't substract from `totalSupply`.
- People can still establish a consensus for token burning, like transfer to `0x000000000000000000000000000000000000dEaD`, but unlike `address(0)`, this is less likely to be accidental

kingster-will:
prevent send wip to zero address or wip contract as well self approval.

issue: resolves #283

### DIFF
--- a/contracts/src/token/WIP.sol
+++ b/contracts/src/token/WIP.sol
@@ -13,6 +13,8 @@ contract WIP is ERC20 {
     error IPTransferFailed();
     /// @notice emitted when an invalid transfer recipient is detected
     error InvalidTransferReceiver();
+    /// @notice emitted when an invalid transfer spender is detected
+    error InvalidTransferSpender();
 
     /// @notice triggered when IP is deposited in exchange for WIP
     receive() external payable {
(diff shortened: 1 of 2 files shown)


limengformal (approved):

contracts/src/token/WIP.sol
```diff
@@ -11,6 +11,10 @@ contract WIP is ERC20 {
     event Withdrawal(address indexed to, uint amount);
     /// @notice emitted when a transfer of IP fails
     error IPTransferFailed();
+    /// @notice emitted when an invalid transfer recipient is detected
+    error InvalidTransferReceiver();
```
Ramarti, line 15: There is an EIP in last call to make standard ERC20 transfer errors
https://eips.ethereum.org/EIPS/eip-6093
```
error ERC20InvalidReceiver(address receiver);
error ERC20InvalidSender(address sender);
error ERC20InvalidSender(address sender);
...
```

kingster-will: renamed the error according to the EIP.

Ramarti (approved):
LGTM

0xHansLee (approved):

PR #291, opened 2024-10-23
feat(params): combine singularity height into one

0xHansLee:
Combined all three singularity height param into one of staking keeper param.

issue: none


0xHansLee:
The version of cosmos-sdk will be updated after new release.

edisonz0718 (approved):

Issue #319, opened 2024-10-25
Odyssey devnet halt


lucas2brh:
## Description and context

The Odyssey devnet halted several hours ago during testing. The test involved a scenario where a delegator repeatedly delegates tokens to multiple validators and then withdraws rewards.

[Test case](https://github.com/storyprotocol/l1-tests/pull/231/files#diff-33492a17a900c3bf98ef3d61e910fea16ffeaf86784e04cbe99128c1908b7b48)

Log
```
	
24-10-25 08:42:18.680 ERRO CONSENSUS FAILURE!!!                     module=consensus err="failed to apply block; error process reward withdrawals: enqueue eligible reward withdrawals: spendable balance 0stake is smaller than 488042902stake: insufficient funds"
2024-10-25 16:42:18.681	
  │ github.com/cometbft/cometbft/consensus.(*State).enterCommit(0xc001714e08, 0x8c88, 0x0)
2024-10-25 16:42:18.681	
  │ github.com/cometbft/cometbft/consensus.(*State).enterCommit.func1()
2024-10-25 16:42:18.681	
  │ \t/home/runner/go/pkg/mod/github.com/cometbft/cometbft@v0.38.9/consensus/state.go:1682 +0x2e8
2024-10-25 16:42:18.681	
  │ github.com/cometbft/cometbft/consensus.(*State).receiveRoutine.func2()
2024-10-25 16:42:18.681	
24-10-25 08:42:18.680 ERRO error in proxyAppConn.FinalizeBlock      module=state err="process reward withdrawals: enqueue eligible reward withdrawals: spendable balance 0stake is smaller than 488042902stake: insufficient funds" stacktrace="[errors.go:189 errors.go:204 send.go:278 send.go:210 keeper.go:302 withdraw.go:309 withdraw.go:85 abci.go:40 module.go:145 module.go:804 app.go:170 baseapp.go:792 abci.go:819 abci.go:884 cmt_abci.go:44 abci.go:99 local_client.go:185 app_conn.go:104 execution.go:224 execution.go:202 state.g
```

## Steps to reproduce

1.
2.
3.

tyitang:
verified on devnet, fixed

PR #321, opened 2024-10-26
fix(evmstaking): fix double spent issue of claimed reward

0xHansLee:
With multiple delegations from a delegator, the claimed reward was withdrawn multiple. Fix the issue by processing eligible reward withdrawal.

issue: #319

### DIFF
--- a/client/x/evmstaking/keeper/withdraw.go
+++ b/client/x/evmstaking/keeper/withdraw.go
@@ -21,13 +21,6 @@ import (
 	"github.com/piplabs/story/lib/log"
 )
 
-type RewardWithdrawal struct {
-	DelegatorAddress string
-	ValidatorAddress string
-	WithdrawalEntry  types.Withdrawal
-	RemainedBalance  uint64
-}
-
 func (k Keeper) ProcessUnbondingWithdrawals(ctx context.Context, unbondedEntries []stypes.UnbondedEntry) error {
 	log.Debug(ctx, "Processing mature unbonding delegations", "count", len(unbondedEntries))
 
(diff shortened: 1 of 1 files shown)


client/x/evmstaking/keeper/withdraw.go
```diff
@@ -239,92 +168,148 @@ func (k Keeper) ExpectedRewardWithdrawals(ctx context.Context) ([]RewardWithdraw
...
+		return err
+	}
 
-		// Withdraw delegation rewards.
-		delAddr := sdk.MustAccAddressFromBech32(rewardWithdrawals[i].DelegatorAddress)
-		delRewards, err := k.distributionKeeper.WithdrawDelegationRewards(ctx, delAddr, valAddr)
+	// if it is self-delegation, add commission
+	if delegation.DelegatorAddress == valAccAddr.String() {
+		// Get validator commissions.
+		valCommission, err := k.distributionKeeper.GetValidatorAccumulatedCommission(ctx, valAddr)
```
jdubpark, line 212: Instead of getting the validator commission in each iteration of delegation, can we pass it in as an argument to this function? The validator commission is constant per validator during the delegation loop, so it'd be more efficient to retrieve it only once and reduce reads.

client/x/evmstaking/keeper/withdraw.go
```diff
@@ -239,92 +168,148 @@ func (k Keeper) ExpectedRewardWithdrawals(ctx context.Context) ([]RewardWithdraw
...
-		err = k.bankKeeper.SendCoinsFromAccountToModule(ctx, delAddr, types.ModuleName, delRewards)
-		if err != nil {
-			return err
-		}
-		err = k.bankKeeper.BurnCoins(ctx, types.ModuleName, delRewards)
-		if err != nil {
-			return err
+	// if total reward is greater than or equal to min reward withdrawal amount, enqueue the reward withdrawal
+	if totalReward >= minRewardWithdrawalAmount {
+		if err := k.EnqueueRewardWithdrawal(ctx, delegation.DelegatorAddress, valAddr.String(), claimedReward); err != nil {
```
jdubpark, line 232: This switches from calling `EnqueueEligibleRewardWithdrawal` after calculating all eligible reward withdrawals (all iterations of delegations) to calling `EnqueueRewardWithdrawal` in each iteration of the delegations. 

Wondering if this breaks any assumption in each iteration's calculation (e.g. we might expect the validator's balance to remain static throughout the loop iterations), or if no assumptions are broken with this new place for logic execution.

0xHansLee: Basically, in CL side, there is no asset unbonded (so transferrable) as far as I know, because staking related actions are started from EL side and unbonded assets are burnt when undelegated. The only case that CL account has asset in bank keeper is that when the delegator makes another delegation under the same validator. (by settling the previous rewards and creating a new starting info for reward).

If you have any cases which could breaks any assumptions in your mind, please lmk. Or any other suggestions on handling this would be great.

client/x/evmstaking/keeper/withdraw.go
0xHansLee: agreed to @ezreal1997.

client/x/evmstaking/keeper/withdraw.go
jdubpark: Thanks, looks good for now.

jdubpark (approved):

PR #328, opened 2024-10-30
feat(x/evmengine): process current block events during finalizeblock after forkchoice updated

jdubpark:
addresses omni-network/omni/issues/2106; event ordering is deterministic on the same payload received via p2p.

issue: none


0xHansLee (approved):
LGTM

client/x/evmengine/keeper/abci_internal_test.go
```diff
@@ -538,10 +538,6 @@ func assertExecutablePayload(t *testing.T, msg sdk.Msg, ts int64, blockHash comm
 	require.Equal(t, payload.FeeRecipient, validatorAddr)
 	require.Empty(t, payload.Withdrawals)
 	require.Equal(t, payload.Number, height)
-
-	// require.Len(t, executionPayload.PrevPayloadEvents, 1)
-	// evmLog := executionPayload.PrevPayloadEvents[0]
-	// require.Equal(t, evmLog.Address, zeroAddr.Bytes())
```
limengformal, line 544: Are they not needed anymore?

client/x/evmengine/keeper/msg_server.go
```diff
@@ -151,15 +151,13 @@ func (s msgServer) ExecutionPayload(ctx context.Context, msg *types.MsgExecution
...
-	if err := s.ProcessUpgradeEvents(ctx, payload.Number-1, msg.PrevPayloadEvents); err != nil {
-		return nil, errors.Wrap(err, "deliver upgrade-related event logs")
+	events, err := s.evmEvents(ctx, payload.BlockHash)
+	if err != nil {
+		return nil, errors.Wrap(err, "fetch evm events")
 	}
-	if err := s.ProcessUbiEvents(ctx, payload.Number-1, msg.PrevPayloadEvents); err != nil {
-		return nil, errors.Wrap(err, "deliver ubi-related event logs")
+
+	if err := s.deliverEvents(ctx, payload.Number, events); err != nil {
```
limengformal, line 159: If we change from processing payload.Number instead of payload.Number - 1 (i assume it is block height), will there be one block of data missing during upgrade?

client/x/evmengine/keeper/msg_server_internal_test.go
```diff
@@ -208,36 +202,35 @@ func Test_msgServer_ExecutionPayload(t *testing.T) {
...
-				return []*types.EVMEvent{{
-					Address: nil, // nil address
-					Topics:  [][]byte{types.SoftwareUpgradeEvent.ID.Bytes()},
-					Data:    data,
-				}}
-			},
-			expectedError: "deliver upgrade-related event logs",
+			expectedError: "deliver staking-related event logs",
 		},
+		//{
```
limengformal, line 208: I assume these commented out tests will be reenabled in the future PRs?

client/x/evmengine/types/tx.proto
```diff
@@ -20,7 +20,6 @@ message MsgExecutionPayload {
   option (cosmos.msg.v1.signer) = "authority";
   string            authority           = 1;
   bytes             execution_payload   = 2;
-  repeated EVMEvent prev_payload_events = 3;
```
0xHansLee, line 23: This field should be reserved.

client/x/evmengine/keeper/msg_server.go
jdubpark: Yes, the linked issue specifies:
> Note that on the network upgrade height itself, we should process both legacy previous block events as well as current block events. Otherwise we will miss events for upgrade height-1.

PR #339, opened 2024-11-05
fix(evmstaking): prevent partial state on failure txn

ezreal1997:
use `CachedContext` to prevent partial state on failure txn

issue: none

### DIFF
--- a/client/x/evmstaking/keeper/keeper_test.go
+++ b/client/x/evmstaking/keeper/keeper_test.go
@@ -1,13 +1,9 @@
 package keeper_test
 
 import (
-	"context"
-	"math/big"
-	"testing"
 	"time"
 
 	"cosmossdk.io/log"
-	sdkmath "cosmossdk.io/math"
 	"cosmossdk.io/store"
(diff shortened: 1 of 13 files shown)


edisonz0718 (approved):

client/x/evmstaking/keeper/keeper_test.go
```diff
@@ -199,6 +180,17 @@ func (s *TestSuite) TestValidatorAddressCodec() {
 	require.NoError(err)
 }
 
+//nolint: dupword // test code
```
0xHansLee, line 183: Instead of commenting out all test cases, I think we can just comment out the case of `pass: process CreateValidatorEvent`, which is the case that we cannot mock the keeper function expectedly.

0xHansLee: And let's find a way to mock the function which uses cached context internally.

ezreal1997: yeah agree, and already uncomment the whole Test func

0xHansLee (approved):

PR #344, opened 2024-11-06
fix: update default blockchain explorer URL in story validator create command

redwest88:
### Summary
This commit updates the default blockchain explorer URL in the `story validator create` command. The previous URL was incorrect, so it's now set to https://odyssey-testnet-explorer.storyscan.xyz to direct users to the correct explorer.


0xHansLee:
With launching Odyssey testnet, the URL works now. So close this PR.

PR #348, opened 2024-11-06
feat(docker): add dockerfile and workflow

wo-o:
Dear team, I have created dockerfile and worflow to make story image.
The REPOSITORY_URI environment and GITHUB_TOKEN secret needed to be set up to run the workflow.

issue: none

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,37 @@
+# Support setting various labels on the final image
+ARG COMMIT=""
+ARG VERSION=""
+ARG BUILDNUM=""
+
+# Build Geth in a stock Go builder container
+FROM golang:1.22-alpine as builder
+
+# Set the Current Working Directory inside the container
+WORKDIR /story
+
+# Copy go.mod and go.sum files
+COPY go.mod go.sum ./
+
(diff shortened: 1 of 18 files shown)


Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+ADD . /story/
+RUN go build -o story ./client
+
+# Pull Geth into a second stage deploy alpine container
+FROM alpine:latest
+
+RUN apk add --no-cache ca-certificates
+COPY --from=builder /story/story /usr/local/bin/
+
+EXPOSE 8545 8546 30303 30303/udp
```
ezreal1997, line 27: these ports seem to be the port of `geth`? maybe we need to expose the ports used by `story`

ezreal1997 (approved):

Dockerfile
0xHansLee: It could be. On CL side, 26656 for p2p and 26657 for rpc. In addition, 1317 (by default) for API should be opened for providing API service.

wo-o: I changed expose port. Thanks!

PR #358, opened 2024-11-14
fix(contracts): add and remove operator refactor

### LINKED ISSUE #302: M-1 Discrepancy between operator management in smart contract and L1
## Description and context
The smart contracts contain a function removeOperator that removes a specified operator (https://github.com/piplabs/story/blob/f1b194ccf1ce4c64239dab3393f48a0c72b20e69/contracts/src/protocol/IPTokenStaking.sol#L179):
```go
    /// @notice Removes an operator for a delegator.
    /// @param uncmpPubkey 65 bytes uncompressed secp256k1 public key.
    /// @param operator The operator address to remove.
    function removeOperator(
        bytes calldata uncmpPubkey,
        address operator
    ) external verifyUncmpPubkeyWithExpectedAddress(uncmpPubkey, msg.sender) {
        emit RemoveOperator(uncmpPubkey, operator);
    }
```

However, the `operator` argument of the emitted event is completely ignored when this event is processed (https://github.com/piplabs/story/blob/46dabe3131eb762c0c827fa3aa3eff47ec95f011/client/x/evmstaking/keeper/delegator_address.go#L85): 
```go
func (k Keeper) ProcessRemoveOperator(ctx context.Context, ev *bindings.IPTokenStakingRemoveOperator) error {
    delCmpPubkey, err := UncmpPubKeyToCmpPubKey(ev.UncmpPubkey)
    if err != nil {
        return errors.Wrap(err, "compress depositor pubkey")
    }
    depositorPubkey, err := k1util.PubKeyBytesToCosmos(delCmpPubkey)
    if err != nil {
        return errors.Wrap(err, "depositor pubkey to cosmos")
    }

    depositorAddr := sdk.AccAddress(depositorPubkey.Address().Bytes())

    if err := k.DelegatorOperatorAddress.Remove(ctx, depositorAddr.String()); err != nil {
        return errors.Wrap(err, "delegator operator address map remove")
    }

    return nil
}
```
The code will just remove the current operator.

A similar discrepancy also exists for `addOperator`. The smart contract mentions that an operator is added (which makes sense based on the function name).
```go
    /// @notice Adds an operator for a delegator.
    /// @param uncmpPubkey 65 bytes uncompressed secp256k1 public key.
    /// @param operator The operator address to add.
    function addOperator(
        bytes calldata uncmpPubkey,
        address operator
    ) external payable verifyUncmpPubkeyWithExpectedAddress(uncmpPubkey, msg.sender) chargesFee {
        emit AddOperator(uncmpPubkey, operator);
    }
```
However, in the L1 code, there can only be ever one operator per delegator (stored in the k.DelegatorOperatorAddress map) and the current operator is overwritten whenever a new one is added.

## Solution recommendation
If multiple operators should be supported, we recommend changing the L1 code and data structures to store all operators. If not, we recommend changing the smart contracts (function name + documentation for `addOperator`, removal of unnecessary event for `removeOperator`).

Ramarti:
Refactor operator methods to reflect there is only 1 operator to set and unset

issue: fixes #302


edisonz0718 (approved):

0xHansLee (commented):
LGTM. Could you rename flags and cmd/validator.go as well?

0xHansLee (approved):
> Could you rename flags and cmd/validator.go as well?

I will handle this. LGTM

PR #369, opened 2024-11-19
feat(cli): separate unjail and unjailOnBehalf cmd

0xHansLee:
Separate unjail command into `unjail` and `unjailOnBehalf`. 

issue: none


client/cmd/validator.go
```diff
@@ -710,6 +732,35 @@ func unstakeOnBehalf(ctx context.Context, cfg unstakeConfig) error {
...
+
+	var unjailFee *big.Int
+	err = cfg.ABI.UnpackIntoInterface(&unjailFee, "fee", result)
+	if err != nil {
+		return errors.Wrap(err, "failed to unpack unjailFee")
+	}
+
+	fmt.Printf("Unjail fee: %s\n", unjailFee.String())
+
+	_, err = prepareAndExecuteTransaction(ctx, &cfg.baseConfig, "unjailOnBehalf", unjailFee, uncompressedValidatorPubKeyBytes, []byte{})
```
edisonz0718, line 753: We should call unjail instead of unjailOnBehalf here

client/cmd/validator.go
```diff
@@ -710,6 +732,35 @@ func unstakeOnBehalf(ctx context.Context, cfg unstakeConfig) error {
...
+	if err != nil {
+		return err
+	}
+
+	fmt.Println("Validator successfully unjailed on behalf of validator!")
```
edisonz0718, line 758: The log message here should be unjail, not unjail on behalf

client/cmd/validator.go
0xHansLee: Thanks for pointing out. It was missed. Applied in [591a551](https://github.com/piplabs/story/pull/369/commits/591a551140e52fb636c60c214e165fe68984f2df).

ezreal1997 (approved):

edisonz0718 (approved):

PR #370, opened 2024-11-19
fix(evmstaking): enhance staking events error handling

ezreal1997:
enhance staking events error handling

issue: none


0xHansLee (approved):
qq: we don't need cometbft API anymore?

ezreal1997:
> qq: we don't need cometbft API anymore?

There's Comet RPC on CometBFT node to query CometBFT state, we can use that.

For here it was added by mistake, we should remove it.

edisonz0718 (approved):

PR #377, opened 2024-11-25
fix(api): enhance api server security

ezreal1997:
enhance api server security

issue: none


edisonz0718:
Can we make the timeout configurable? We can include this in a separate PR

edisonz0718 (approved):

ezreal1997:
> Can we make the timeout configurable? We can include this in a separate PR

good point, we could have related config  under  api section in `story.toml`

0xHansLee (approved):

PR #380, opened 2024-11-26
fix(contracts): various fixes for audit issues

Ramarti:
- TRST-R-13: Remove unused Predeploys.isPredeploy() and isActivePredeploy()
- TRST-R-14 Update the documentation around the WIP token deployment and predeploy namespaces
- TRST-R-4 Check zero amounts after the rounding in IPTokenStaking contract
- TRST-L-1 Lack of UBI token distribution segregation might prevent validators from claiming their entitled rewards

issue: none


0xHansLee (approved):
LGTM. Please regenerate bindings as well before merging.

limengformal:
As discussed with @edisonz0718 , we will not fix https://github.com/piplabs/story/issues/393 in odyssey since it is a name change. We will apply the fix in mainnet version.

contracts/src/protocol/UBIPool.sol
```diff
@@ -88,5 +93,6 @@ contract UBIPool is
         validatorUBIAmounts[distributionId][validatorUncmpPubkey] = 0;
         (bool success, ) = msg.sender.call{ value: amount }("");
         require(success, "UBIPool: failed to send UBI");
+        totalPendingClaims -= amount;
```
jdubpark, line 96: In `setUBIDistribution`, we don't check if the sum of `amounts` equals `totalPendingClaims`, so it's possible that `totalPendingClaims` might underflow (if the sum of `amounts` > `totalPendingClaims`).

Ramarti: When we `setUBIDistribution`, we check
```
// enough balance in contract
require(totalUBI + totalPendingClaims <= address(this).balance, "UBIPool: not enough balance");
// add to totalPendingClaims the ubi being distributed
totalPendingClaims += totalUBI;
        uint256 accAmount;
        for (uint256 i = 0; i < amounts.length; i++) {
            //...
            accAmount += amounts[i];
        }
        // Checking that total is actually the sum of individual amounts.
        require(accAmount == totalUBI, "UBIPool: total amount mismatch");
```

The only place we withdraw is `claimUBI`, where the highlighted code is. There is no way for `amount` to be > `totalPendingClaims`. If it was for some reason, solidity reverts on underflow after 0.8.0 so that claim might be stuck but no underflow.

PR #392, opened 2024-11-30
chore(*): enable optimistic build

### LINKED ISSUE #389: Enable optimistic build
## Description and context
Enable optimistic build for Odyssey. It should ideally reduce the block time. This doesn't require hardfork

## Definition of done
Change added and tested

edisonz0718:
Enable optimistic build in Odyssey

issue: close #389

### DIFF
--- a/client/config/config.go
+++ b/client/config/config.go
@@ -74,7 +74,7 @@ var (
 		MinRetainBlocks:    defaultMinRetainBlocks,
 		PruningOption:      pruningtypes.PruningOptionDefault,
 		EVMBuildDelay:      defaultEVMBuildDelay,
-		EVMBuildOptimistic: false,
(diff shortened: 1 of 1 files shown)


LeoHChen:
What's the implication?  Why it wasn't working before?  I want to make sure we solved all the risk.

edisonz0718:
> What's the implication? Why it wasn't working before? I want to make sure we solved all the risk.

We disabled it in Iliad since we thought the function cannot get ctx from cometbft when building optimistic build. However this is not the case. We've tested in in devnet environment that it works and reduce block time by around 0.5s.

0xHansLee (approved):

ezreal1997 (approved):

jdubpark (approved):

PR #397, opened 2024-12-02
fix(prouter): add tx number check

### LINKED ISSUE #376: Add tx count check in processProposal
## Description and context
Our current design guarantees that only 1 cosmos tx will be added to a block, except the block 1. So we should add this check to process proposal phase to make sure no malicious actors can attack us by sending more messages

## Suggested solution
reference PR: https://github.com/omni-network/omni/pull/2529

## Definition of done
Implement and test the change

edisonz0718:
Add the check to make sure the node receive the correct number of txs.

Tested in local testnet

issue: close #376

### DIFF
--- a/client/app/prouter.go
+++ b/client/app/prouter.go
@@ -40,19 +40,27 @@ func makeProcessProposalHandler(router *baseapp.MsgServiceRouter, txConfig clien
 		defer timeoutCancel()
 		ctx = ctx.WithContext(timeoutCtx)
 
-		// Ensure the proposal includes quorum vote extensions (unless first block).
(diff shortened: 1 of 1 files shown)


ezreal1997 (approved):

LeoHChen:
How to re-enable all the tests? like the original PR?

0xHansLee (approved):

edisonz0718:
> How to re-enable all the tests? like the original PR?

We will add unit tests as discussed. But the PR is tested in our local testnet with 4 validator nodes and 1 bootnode

PR #400, opened 2024-12-04
fix(events): align event processing with smart contract changes

### LINKED ISSUE #393: add new operate event handlers
## Description and context
To address [issue](https://github.com/piplabs/story/issues/302), Operator related events will be updated on EL side. We will need to update CL side event handler before that.

## Suggested solution
To make CL backward compatible, add new event handler to handle new events.

ezreal1997:
align event processing with smart contract changes

issue: closes #393


client/x/evmstaking/keeper/delegator_address.go
```diff
@@ -148,7 +148,6 @@ func (k Keeper) ProcessRemoveOperator(ctx context.Context, ev *bindings.IPTokenS
 				types.EventTypeRemoveOperatorFailure,
 				sdk.NewAttribute(types.AttributeKeyBlockHeight, strconv.FormatInt(sdkCtx.BlockHeight(), 10)),
 				sdk.NewAttribute(types.AttributeKeyDelegatorUncmpPubKey, hex.EncodeToString(ev.UncmpPubkey)),
-				sdk.NewAttribute(types.AttributeKeyOperatorAddress, ev.Operator.Hex()),
```
edisonz0718, line 151: Why do we remove this attribute

edisonz0718 (commented):
I think we need to make this backward compatible since it's tricky to upgrade CL and contract together.

jdubpark:
> #393

I agree, it requires precise timing of smart contract upgrade to follow the updated CL.

Alternatively we could do this:
1. Update the smart contract to emit two logs in each operator function, ie. emit `AddOperator` & `SetOperator` and `RemoveOperator` & `UnsetOperator`.
2. Update the CL code.
3. Nodes that have not upgraded will pick old logs (Add/Remove) and nodes that have upgraded will pick new logs (Set/Unset).
4. At some point, update the smart contract to remove the `AddOperator` and `RemoveOperator` event logs.

client/x/evmstaking/keeper/delegator_address.go
ezreal1997: new event does not include `operator_address`, we could get from our KVS but imo it's useless as we only support single operator address and it's failure log

ezreal1997:
> I think we need to make this backward compatible since it's tricky to upgrade CL and contract together.

Anyway, we need to notify the community that we will deprecate `AddOperator` and `RemoveOperator`, so I prefer a simpler way:
1. Include this change in the next hardfork.
2. Then we notify the community 1) old contract funcs to be deprecated and occurrence time is the next hardfork time, 2) new contract funcs to be supported and the occurrence time is xxx blocks/hours right after hardfork.

cc @edisonz0718 @jdubpark

edisonz0718:
> > I think we need to make this backward compatible since it's tricky to upgrade CL and contract together.
> 
> Anyway, we need to notify the community that we will deprecate `addOperator` and `removeOperator` func, so I prefer a simpler way:
> 
> 1. Include this change in the next hardfork.
> 2. Then we notify the community 1) old contract funcs to be deprecated and occurrence time is the next hardfork time, 2) new contract funcs to be supported and the occurrence time is xxx blocks/hours right after hardfork.
> 
> cc @edisonz0718 @jdubpark

This will cause a service disruption if someone wants to use operator functions between CL hardfork and contract upgrade. I think a simpler approach is to just make CL backward compatible. Add the new operator functions instead of replacing. So that both types of event can be processed.

ezreal1997:
> > > I think we need to make this backward compatible since it's tricky to upgrade CL and contract together.
> > 
> > 
> > Anyway, we need to notify the community that we will deprecate `addOperator` and `removeOperator` func, so I prefer a simpler way:
> > 
> > 1. Include this change in the next hardfork.
> > 2. Then we notify the community 1) old contract funcs to be deprecated and occurrence time is the next hardfork time, 2) new contract funcs to be supported and the occurrence time is xxx blocks/hours right after hardfork.
> > 
> > cc @edisonz0718 @jdubpark
> 
> This will cause a service disruption if someone wants to use operator functions between CL hardfork and contract upgrade. I think a simpler approach is to just make CL backward compatible. Add the new operator functions instead of replacing. So that both types of event can be processed.

If we plan to keep both operator functions until `Odyssey` deprecated, that makes sense. @Ramarti plz add the old operator func back.

0xHansLee:
Yeah, I also agree to @edisonz0718. We can make CL backward compatible for the contract upgrade for now, and deprecate the old events after upgrading the contract.

ezreal1997:
My point is if we do not keep old operator functions, it's already a non-compatible change on user side, so it's equivalent if CL is backward compatible or not.

If we want to make backward compatible, we need following steps:
1. upgrade CL to handler both old and new events
2. upgrade contract to keep both old and new operator functions
3. ... one month for eco projects to make the change ...
4. upgrade the contract to remove the old operator functions
5. upgrade CL to remove the processing logic of old events

limengformal:
As discussed with @edisonz0718 , we will not fix https://github.com/piplabs/story/issues/393 in odyssey since it is a name change. We will apply the fix in mainnet version.

PR #401, opened 2024-12-04
feat(upgrade): upgrade handler v0.14.0

ezreal1997:
upgrade handler for Cosmos (HAL-02) Key typo may allow store corruption

issue: none

### DIFF
--- a/client/app/upgrades/v0_13_2/upgrades.go
+++ b/client/app/upgrades/v0_13_2/upgrades.go
@@ -1,0 +1,59 @@
+//nolint:revive,stylecheck // version underscores
+package v0_13_2
+
+import (
+	"context"
+
+	upgradetypes "cosmossdk.io/x/upgrade/types"
+
+	"github.com/cosmos/cosmos-sdk/types/module"
+
+	"github.com/piplabs/story/client/app/keepers"
+	"github.com/piplabs/story/lib/errors"
+	"github.com/piplabs/story/lib/log"
+)
+
+func CreateUpgradeHandler(
+	mm *module.Manager,
+	configurator module.Configurator,
+	keepers *keepers.Keepers,
+) upgradetypes.UpgradeHandler {
+	return func(ctx context.Context, _ upgradetypes.Plan, vm module.VersionMap) (module.VersionMap, error) {
+		log.Info(ctx, "Starting module migrations...")
+
+		vm, err := mm.RunMigrations(ctx, configurator, vm)
+		if err != nil {
+			return vm, errors.Wrap(err, "run migrations")
+		}
+
+		log.Info(ctx, "Getting existing UnbondingID...")
+		unbondingID, err := keepers.StakingKeeper.GetOldUnbondingID(ctx)
+		if err != nil {
+			return vm, errors.Wrap(err, "failed to get existing UnbondingID")
+		}
+		log.Info(ctx, "Existing UnbondingID", "UnbondingID", unbondingID)
+
+		log.Info(ctx, "Setting new UnbondingID...")
+		if err := keepers.StakingKeeper.SetUnbondingID(ctx, unbondingID); err != nil {
+			return vm, errors.Wrap(err, "failed to set new UnbondingID")
+		}
+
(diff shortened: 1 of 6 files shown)


limengformal:
@ezreal1997 Is it going to be a hard fork upgrade? If yes, let's increase minor version and use v0.14.0

client/app/upgrades/v0_13_2/upgrades.go
```diff
@@ -0,0 +1,59 @@
...
+			return vm, errors.Wrap(err, "failed to get new UnbondingID")
+		}
+		log.Info(ctx, "New UnbondingID", "UnbondingID", newUnbondingID)
+
+		if newUnbondingID != unbondingID {
+			return vm, errors.New("new UnbondingID does not match existing UnbondingID")
+		}
+
+		log.Info(ctx, "Removing old UnbondingID...")
+		// TODO: Remove old UnbondingID
```
jdubpark, line 53: cross-ref piplabs/cosmos-sdk#29

ezreal1997:
> @ezreal1997 Is it going to be a hard fork upgrade? If yes, let's increase minor version and use v0.14.0

yeah, it's hard fork upgrade. version updated in piplabs/story@76871e63e54642c0efcc9449824df75139b11c79

client/app/upgrades/v0_13_2/upgrades.go
ezreal1997: updated in piplabs/story@76871e63e54642c0efcc9449824df75139b11c79

jdubpark (approved):

ezreal1997:
> I think it would be better to increase ConsensusVersion of our staking module as well (https://github.com/piplabs/cosmos-sdk/blob/bf84b223e3a00519b6681062de3fb5bc74db78f5/x/staking/module.go#L36).

There's only a key typo fix on cosmos this time, actually we should increase ConsensusVersion when we made the breaking changes on staking module previously. It's ok to make the version increment together with this path.

ezreal1997:
> > I think it would be better to increase ConsensusVersion of our staking module as well (https://github.com/piplabs/cosmos-sdk/blob/bf84b223e3a00519b6681062de3fb5bc74db78f5/x/staking/module.go#L36).
> 
> There's only a key typo fix on cosmos this time, actually we should increase ConsensusVersion when we made the breaking changes on staking module previously. It's ok to make the version increment together with this patch.

@0xHansLee Consensus version change is related to the migration and upgrade logic on cosmos side, maybe we just keep that untouched as the migration/upgrade happens on `story`, wdyt?

0xHansLee:
> > > I think it would be better to increase ConsensusVersion of our staking module as well (https://github.com/piplabs/cosmos-sdk/blob/bf84b223e3a00519b6681062de3fb5bc74db78f5/x/staking/module.go#L36).
> > 
> > 
> > There's only a key typo fix on cosmos this time, actually we should increase ConsensusVersion when we made the breaking changes on staking module previously. It's ok to make the version increment together with this patch.
> 
> @0xHansLee Consensus version change is related to the migration and upgrade logic on cosmos side, maybe we just keep that untouched as the migration/upgrade happens on `story`, wdyt?

_The consensus version serves as the breaking change version of the module._ (from [cosmos-sdk docs](https://docs.cosmos.network/main/learn/advanced/upgrade#tracking-module-versions)).

This upgrade leads to migration of store so I think we need to increase the consensus version. And even this is on staking module in cosmos-sdk, since we made change on the module, so we need to manage our own version I think. 

IMO, once we have some diff from the original one, meaning that fork the module, we need to have our own version.

0xHansLee:
> @0xHansLee Consensus version change is related to the migration and upgrade logic on cosmos side, maybe we just keep that untouched as the migration/upgrade happens on story, wdyt?

Ah, I think I got your point. I think we should have implemented the migration handler in our staking module and increase the consensus version. And then, when we bump up the cosmos-sdk version, the staking module would be upgraded by calling run `RunMigrations()`. BTW, I think it's ok to merge this handler at this time.

0xHansLee (approved):

PR #405, opened 2024-12-05
feat(withdrawal): hardfork needed distinguish withdrawal types

### LINKED ISSUE #395: Use Validator field in EngineAPI withdrawals to distinguish unstake and reward distribution
## Description and context
Currently we have two types of withdrawals. One is unstaking that is triggered by users. The other is reward distribution which happens automatically. From the outside, there is no way to know which type of withdrawal it is. We should find a way to distinguish them. This will help indexers and other apps to show correct information to the users.

## Suggested solution
Since the Validator field is currently set to 0 and not used by EL, we can use this field to mark the type of withdrawals. 0 for unstaking, 1 for rewards.

ezreal1997:
Use `Validator` field in EngineAPI withdrawals to distinguish unstake, reward and ubi distribution

issue: closes #395


0xHansLee (commented):
This might require hardfork, right? Then, let's edit the title of PR to notice that upgrade handler is required for this.

0xHansLee (approved):
If the upgrade handler for this change is implemented in a separate PR, looks good to me.

jdubpark (approved):

ezreal1997:
> This might require hardfork, right? Then, let's edit the title of PR to notice that upgrade handler is required for this.

hardfork needed, but we don't need any data migration since proto3 is both forward and backward compatible

edisonz0718 (approved):

client/x/evmstaking/keeper/withdraw.go
```diff
@@ -60,9 +60,10 @@ func (k Keeper) ProcessUnbondingWithdrawals(ctx context.Context, unbondedEntries
 			uint64(sdk.UnwrapSDKContext(ctx).BlockHeight()),
 			delEvmAddr,
 			entry.Amount.Uint64(),
+			types.WithdrawalType_WITHDRAWAL_TYPE_UNSTAKE,
 		))
 		if err != nil {
-			return errors.Wrap(err, "add unbonding withdrawal to queue")
```
edisonz0718, line 65: nit: unbonded may be a better word to use here

PR #406, opened 2024-12-06
chore(gomod): rollback cometbft to v0.38.9

0xHansLee:
After upgrading cometbft to v0.38.12, we encountered `sent next PEX request too soon` error.
Before fixing this issue, I would like to rollback the cometbft to v0.38.9 (go-ethereum and btcec as well since there are related breaking changes from v0.38.9 to v0.38.12).

issue: none


go.mod
```diff
@@ -23,7 +23,7 @@ require (
 	github.com/cosmos/cosmos-sdk v0.50.7
 	github.com/cosmos/gogoproto v1.7.0
 	github.com/decred/dcrd/dcrec/secp256k1/v4 v4.3.0
-	github.com/ethereum/go-ethereum v1.14.8
```
edisonz0718, line 26: Do we also need to rollback geth version?

0xHansLee: Yes, because of the compatibility issue of btcec. 
cometbft v0.38.9 uses v2.2.0 of btcec, but go-ethereum v1.14.8 uses v2.3.4 of btcec, which is not backward compatible.

jdubpark (approved):

ezreal1997:
do we also need to rollback cosmos-sdk to `v0.50.8`? I remember the reason to upgrade cosmos-sdk to `v0.50.10` is to be compatible with cometBFT `v0.38.12`

0xHansLee:
> do we also need to rollback cosmos-sdk to v0.50.8? I remember the reason to upgrade cosmos-sdk to v0.50.10 is to be compatible with cometBFT v0.38.12

It was (but not required). But it seems v0.50.10 is also compatible with cometbft v0.38.9 (57block tested). So I just revert cometbft since there are some changes you made in cosmos-sdk as well.

ezreal1997 (approved):

PR #409, opened 2024-12-07
fix(evmengine): get next proposer via comet algo

### LINKED ISSUE #388: Update isNextProposer algorithm
## Description and context
Currently the isNextProposer algorithm: round-robin is different from the actual algorithm used by cometbft. We should use the same algo to pick next proposer

## Solution recommendation
Query cometbft to get the correct next proposer. Reference from omni: https://github.com/omni-network/omni/pull/2526/files#diff-5b776d2b7b8e2c9264856050ff385fcf806c6d2d7eca60cc4a54c96065620e24

jdubpark:
use comet's exposed validators functions to check if the local node is the next (predicted) proposer

issue: closes #388

### DIFF
--- a/client/x/evmengine/keeper/keeper_internal_test.go
+++ b/client/x/evmengine/keeper/keeper_internal_test.go
@@ -29,10 +29,9 @@ import (
 
 type args struct {
 	height         int64
-	validatorsFunc func(context.Context, int64) (*cmttypes.ValidatorSet, bool, error)
-	current        int
-	next           int
-	header         func(height int64, address []byte) cmtproto.Header
+	validatorsFunc func(context.Context, int64) (*cmttypes.ValidatorSet, error)
+	incMoreTimes   int32
+	header         func(height int64) cmtproto.Header
 }
 
 func createTestKeeper(t *testing.T) (context.Context, *Keeper) {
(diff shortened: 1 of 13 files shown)


client/x/evmengine/keeper/keeper_internal_test.go
```diff
@@ -292,81 +287,43 @@ func TestKeeper_isNextProposer(t *testing.T) {
...
-			name: "is next proposer",
+			name: "not proposer",
 			args: args{
-				height:  height,
-				current: 0,
-				next:    1,
-				header: func(height int64, address []byte) cmtproto.Header {
-					return cmtproto.Header{Height: height, ProposerAddress: address}
+				height:       height,
+				incMoreTimes: 9,
```
0xHansLee, line 299: This argument seems to cause the test to fail sometimes. Why don't we just take a bool whether it's nextProposer or not and set the validator address to something other than the proposer if it's not?

jdubpark: Good point, I've done that in `abci_internal_test.go` but missed this one. Resolved in e35ea5a, if not next proposer, then set `nxtAddr` (next proposer) to zero address (`0x00...00`)

0xHansLee (approved):
LGTM

edisonz0718 (approved):

PR #416, opened 2024-12-11
fix(evmengine): pseudorandom randao

jdubpark:
use pseudorandomness for RANDAO

issue: closes piplabs/trust-story#7

### DIFF
--- a/client/x/evmengine/keeper/abci.go
+++ b/client/x/evmengine/keeper/abci.go
@@ -16,6 +16,7 @@ import (
 	etypes "github.com/ethereum/go-ethereum/core/types"
 
 	"github.com/piplabs/story/client/x/evmengine/types"
+	"github.com/piplabs/story/lib/crypto"
 	"github.com/piplabs/story/lib/errors"
 	"github.com/piplabs/story/lib/log"
 )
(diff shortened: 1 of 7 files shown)


client/x/evmengine/keeper/abci.go
```diff
@@ -280,7 +281,7 @@ func (k *Keeper) startBuild(ctx context.Context, feeRecipient common.Address, wi
 
 	attrs := &engine.PayloadAttributes{
 		Timestamp:             ts,
-		Random:                head.Hash(), // We use head block hash as randao.
+		Random:                crypto.RandomHash(),
```
ezreal1997, line 284: shall we guarantee the deterministic based on the proposer?

jdubpark: What would be required for that? 

I think some kind of determinism makes randomness not random enough, if that makes sense

Ramarti: This is how it's solved in Prism (ETH CL client)
https://github.com/prysmaticlabs/prysm/blob/develop/beacon-chain/core/blocks/randao.go

Value is derivated from beacon state and validator's key.

https://eips.ethereum.org/EIPS/eip-4399
> Obviously, historical randomness provided by any decentralized oracle is 100% predictable.
> Make your applications rely on the future randomness with a reasonably high lookahead

Use of RANDAO is not very extended for applications, I think it's better to have this be a known issue that we document, rather than rushing a solution.

edisonz0718 (commented):
This won't be able to pass the block verification here: https://github.com/piplabs/story/blob/main/client/x/evmengine/keeper/keeper.go#L173

Also agree with @Ramarti  that we shouldn't do a rush fix here. Proper comment and documentation is good enough at this stage. We can provide other randomness options like Chainlink VRF.

PR #425, opened 2024-12-12
feat(contracts): secp256k1 verify on curve

### LINKED ISSUE #375: require that pubkeys are on the secp256k1 curve
## Description and context
We currently don't check if a pubkey is on secp256k1. Only check if it's a valid length

## Suggested solution
Add the check to the contract

## Definition of done
Implement and test the change

jdubpark:
checks that the argument (uncompressed) secp256k1 pubkey is on curve

issue: closes #375


ezreal1997 (approved):

Ramarti (approved):
LGTM

PR #426, opened 2024-12-12
fix(*): resolve and close todos

jdubpark:
address open TODOs

issue: piplabs/trust-story#19

### DIFF
--- a/client/app/start.go
+++ b/client/app/start.go
@@ -252,7 +252,7 @@ func makeBaseAppOpts(cfg Config) ([]func(*baseapp.BaseApp), error) {
 	}
 
 	return []func(*baseapp.BaseApp){
-		baseapp.SetOptimisticExecution(),
+		// baseapp.SetOptimisticExecution(), // TODO: research this feature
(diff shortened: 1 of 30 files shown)


client/app/start.go
```diff
@@ -286,7 +252,7 @@ func makeBaseAppOpts(cfg Config) ([]func(*baseapp.BaseApp), error) {
 	}
 
 	return []func(*baseapp.BaseApp){
-		// baseapp.SetOptimisticExecution(), // TODO: Enable this.
```
edisonz0718, line 289: This is an experimental feature and may have impact on things that we don't know. Let's not enable it for now.

edisonz0718: Instead, let's create an issue to track this

0xHansLee (approved):

client/app/start.go
jdubpark: Sure, I got it confused with optimistic building. Disabled in 0498250

edisonz0718 (approved):

PR #443, opened 2024-12-17
feat(genesis): mainnet values in gen alloc and testing

Ramarti:
- Unifies Timelock initialization across chains
- Fixes admin (if `KEEP_TIMELOCK_ADMIN_ROLE = true`)
- Refactor chain ids into utils library
- Sets min delay values in mainnet
- Test upgrade of last Predeploy proxy and scheduling over min delay

issue: none


contracts/script/utils/ChainIds.sol
```diff
@@ -0,0 +1,11 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.23;
+
+library ChainIds {
+    uint256 public constant MAINNET = 1111; // TBD
```
edisonz0718, line 5: To be consistent, let's call it STORY_MAINNET and set id to 1415

0xHansLee (approved):
lgtm

contracts/script/utils/ChainIds.sol
```diff
@@ -0,0 +1,11 @@
...
+    uint256 public constant ILIAD = 1513;
+    uint256 public constant MININET = 1512;
+    uint256 public constant ODYSSEY_DEVNET = 1315;
+    uint256 public constant ODYSSEY_TESTNET = 1516;
+    uint256 public constant LOCAL = 31337;
```
limengformal, line 10: I think local is 1511? @edisonz0718

contracts/script/GenerateAlloc.s.sol
```diff
@@ -217,25 +197,34 @@ contract GenerateAlloc is Script {
         // We deploy this with Create3 because we can't set storage variables in constructor with vm.etch
 
         uint256 minDelay = getTimelockMinDelay();
-        (address[] memory proposers, address[] memory executors, address canceller) = getTimelockControllers();
+        address[] memory proposers = new address[](1);
+        proposers[0] = protocolAdmin;
```
limengformal, line 201: I guess this hasn't been changed to multi-sig yet?

contracts/script/GenerateAlloc.s.sol
```diff
@@ -217,25 +197,34 @@ contract GenerateAlloc is Script {
...
+        address[] memory executors = new address[](1);
+        executors[0] = timelockExecutor;
```
limengformal, line 203: Is it setting it to anyone can execute?

contracts/script/GenerateAlloc.s.sol
```diff
@@ -217,25 +197,34 @@ contract GenerateAlloc is Script {
...
+        address canceller = timelockGuardian;
```
limengformal, line 204: I guess this hasn't been changed to multi-sig yet?

contracts/script/utils/ChainIds.sol
0xHansLee: You are right. For local, chain ID of 1511 is used for now.

Ramarti: Maybe there is a different understanding of local, that's foundry's `anvil` local network, for tests

contracts/script/utils/ChainIds.sol
Ramarti: Done, thanks

contracts/script/utils/ChainIds.sol
Ramarti: I added separated variables for LOCAL and FOUNDRY

contracts/script/GenerateAlloc.s.sol
Ramarti: It allows having address(0) in env files so it's public
https://github.com/piplabs/story/blob/0ac4627edd4d410910a926ca02763e42679aea15/contracts/script/GenerateAlloc.s.sol#L132

contracts/script/GenerateAlloc.s.sol
Ramarti: No need to change anything here, multisigs are identified by EVM address, just need to have the right .env values

contracts/script/GenerateAlloc.s.sol
```diff
@@ -49,87 +50,69 @@ contract GenerateAlloc is Script {
...
 
     /// @notice this call should only be available from Test.sol, for speed
     function disableStateDump() external {
-        require(block.chainid == 31337, "Only for local tests");
+        require(block.chainid == ChainIds.FOUNDRY, "Only for local tests");
         saveState = false;
     }
 
     /// @dev this call should only be available from Test.sol
     function setAdminAddresses(address protocol, address executor, address guardian) external {
```
jdubpark, line 65: FWIW, we could write to internal storage slots directly in tests (but having this function in script doesn't impact performance or anything).

jdubpark (approved):

limengformal (approved):

PR #472, opened 2025-02-04
feat(upgrade): change rewards multiplier for mainnet

0xHansLee:
Add to change reward multiplier of period delegation for Mainnet. 

For Aeneid, the Virgil upgrade has already executed, so this change of upgrade handler do not change the multipliers for Aeneid. That means, anyone can sync blocks from the genesis with the new binary. 

This will only change the multipliers for Mainnet (`chainID` = `story-1`). This assumes that no period delegations will happen until the upgrade height since it will be in the singularity. 

Tested with localnet for the 2 cases
- mocking Aeneid (already upgrade executed, and sync block from the genesis with this new binary)
- mocking Mainnet (change staking params including rewards multiplier)

issue: none

### DIFF
--- a/client/app/upgrades/singularity/virgil/constants.go
+++ b/client/app/upgrades/singularity/virgil/constants.go
@@ -17,7 +17,7 @@ const (
 	// AeneidUpgradeHeight defines the block height at which virgil upgrade is triggered on Aeneid.
 	AeneidUpgradeHeight = 345158
 	// StoryUpgradeHeight defines the block height at which virgil upgrade is triggered on Story.
-	StoryUpgradeHeight = 677886
+	StoryUpgradeHeight = 809988
 )
 
 var Upgrade = upgrades.Upgrade{
--- a/client/app/upgrades/singularity/virgil/constants.go
+++ b/client/app/upgrades/singularity/virgil/constants.go
@@ -1,6 +1,7 @@
 package virgil
 
 import (
+	"cosmossdk.io/math"
 	storetypes "cosmossdk.io/store/types"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
@@ -32,6 +33,12 @@ var Fork = upgrades.Fork{
 	BeginForkLogic: func(_ sdk.Context, _ *keepers.Keepers) {},
 }
 
+type RewardsMultipliers struct {
+	Short  math.LegacyDec
+	Medium math.LegacyDec
+	Long   math.LegacyDec
+}
+
 func GetUpgradeHeight(chainID string) (int64, bool) {
 	switch chainID {
 	case upgrades.AeneidChainID:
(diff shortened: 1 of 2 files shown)


jdubpark (approved):

client/app/upgrades/singularity/virgil/constants.go
```diff
@@ -42,3 +49,22 @@ func GetUpgradeHeight(chainID string) (int64, bool) {
...
+
+func GetRewardsMultipliers(chainID string) RewardsMultipliers {
+	switch chainID {
+	case upgrades.StoryChainID:
+		return RewardsMultipliers{
+			Short:  math.LegacyNewDecWithPrec(11, 1), // 1.1
+			Medium: math.LegacyNewDecWithPrec(15, 1), // 1.5
+			Long:   math.LegacyNewDecWithPrec(20, 1), // 2
+		}
+	default:
```
ezreal1997, line 67: what if we have other devnet/testnet in the future? they may also apply this mainnet default multiplier

client/app/upgrades/singularity/virgil/upgrades.go
```diff
@@ -45,17 +45,24 @@ func CreateUpgradeHandler(
 		}
 
 		log.Info(ctx, "Update staking periods...")
+		newRewardsMultiplier := GetRewardsMultipliers(chainID)
```
ezreal1997, line 48: need to change the `rewards_shares` and `rewards_token` for period_delegation, delegation and validator.

client/app/upgrades/singularity/virgil/constants.go
0xHansLee: I assumed that the default values are current ones for the existing networks. 
For other network in the future, I don't think this upgrade handler is not needed if we set the mainnet default multiplier from the genesis.

client/app/upgrades/singularity/virgil/upgrades.go
0xHansLee: I know what you meant. As I described in the PR body, it is based on the assumption that there is no period delegation until the upgrade height.

ezreal1997 (approved):

client/app/upgrades/singularity/virgil/constants.go
```diff
@@ -16,7 +17,7 @@ const (
 	// AeneidUpgradeHeight defines the block height at which virgil upgrade is triggered on Aeneid.
 	AeneidUpgradeHeight = 345158
 	// StoryUpgradeHeight defines the block height at which virgil upgrade is triggered on Story.
-	StoryUpgradeHeight = 677886
+	StoryUpgradeHeight = 809988
```
ramtinms, line 20: is this change expected for this PR?

limengformal: yes, it is expected. We will allow more time (next Monday) for validators to upgrade

limengformal (approved):

ramtinms (approved):

PR #477, opened 2025-02-13
feat(cli): add update validator commission cli

0xHansLee:
Added a CLI for updating validator commission

issue: none


jdubpark (approved):
Should we fail in cli if % is > 100%? Or under 5% (to maintain error parity with contracts)

0xHansLee:
> Should we fail in cli if % is > 100%? Or under 5% (to maintain error parity with contracts)

OK. Let me add those validation.

0xHansLee:
@jdubpark @ezreal1997 
We do not validate those things which require interaction with contract or RPC for now. Let's improve our CLIs at once in another PR.

PR #479, opened 2025-02-14
feat(cli): add verifications to cli

0xHansLee:
Added some verifications to CLI to prevent potential fund loss.
- create validate: check if the validator does not exist
- stake, unstake, redelegate, unjail, and update commission rate (including onBehalf): check if the validator exist
- update commission rate: check if the new commission rate is valid based on the previous one

issue: none

### DIFF
--- a/client/cmd/flags.go
+++ b/client/cmd/flags.go
@@ -67,7 +67,7 @@ func bindValidatorBaseFlags(cmd *cobra.Command, cfg *baseConfig) {
 	cmd.Flags().StringVar(&cfg.RPC, "rpc", "https://storyrpc.io", "RPC URL to connect to the network")
 	cmd.Flags().StringVar(&cfg.Explorer, "explorer", "https://storyscan.xyz", "URL of the blockchain explorer")
 	cmd.Flags().Int64Var(&cfg.ChainID, "chain-id", 1514, "Chain ID to use for the transaction")
-	cmd.Flags().StringVar(&cfg.API, "api", "", "URL of consensus client API server")
+	cmd.Flags().StringVar(&cfg.StakingAPI, "staking-api", "", "URL of Staking API server")
 }
 
 func bindValidatorCreateFlags(cmd *cobra.Command, cfg *createValidatorConfig) {
@@ -232,11 +232,11 @@ func validateValidatorCreateFlags(ctx context.Context, cmd *cobra.Command, cfg *
 		return errors.Wrap(err, "failed to extract compressed pub key")
 	}
 
-	if cfg.API == "" {
+	if cfg.StakingAPI == "" {
 		return nil
 	}
 
-	exist, err := isValidatorFound(ctx, cfg.API, validatorPubKey)
+	exist, err := isValidatorFound(ctx, cfg.StakingAPI, validatorPubKey)
 	if err != nil {
 		return err
 	}
@@ -278,11 +278,11 @@ func validateValidatorStakeFlags(ctx context.Context, cmd *cobra.Command, cfg *s
 		return errors.Wrap(err, "failed to decode hex-encoded pub key")
 	}
 
-	if cfg.API == "" {
+	if cfg.StakingAPI == "" {
 		return nil
 	}
 
-	exist, err := isValidatorFound(ctx, cfg.API, validatorPubKey)
+	exist, err := isValidatorFound(ctx, cfg.StakingAPI, validatorPubKey)
 	if err != nil {
 		return err
 	}
@@ -308,11 +308,11 @@ func validateValidatorStakeOnBehalfFlags(ctx context.Context, cmd *cobra.Command
 		return errors.Wrap(err, "failed to decode hex-encoded pub key")
 	}
 
-	if cfg.API == "" {
+	if cfg.StakingAPI == "" {
 		return nil
 	}
 
-	exist, err := isValidatorFound(ctx, cfg.API, validatorPubKey)
+	exist, err := isValidatorFound(ctx, cfg.StakingAPI, validatorPubKey)
 	if err != nil {
 		return err
 	}
@@ -338,11 +338,11 @@ func validateValidatorUnstakeFlags(ctx context.Context, cmd *cobra.Command, cfg 
 		return errors.Wrap(err, "failed to decode hex-encoded pub key")
 	}
 
-	if cfg.API == "" {
+	if cfg.StakingAPI == "" {
 		return nil
 	}
 
-	exist, err := isValidatorFound(ctx, cfg.API, validatorPubKey)
+	exist, err := isValidatorFound(ctx, cfg.StakingAPI, validatorPubKey)
 	if err != nil {
 		return err
 	}
@@ -368,11 +368,11 @@ func validateValidatorUnstakeOnBehalfFlags(ctx context.Context, cmd *cobra.Comma
 		return errors.Wrap(err, "failed to decode hex-encoded pub key")
 	}
 
-	if cfg.API == "" {
+	if cfg.StakingAPI == "" {
 		return nil
 	}
 
-	exist, err := isValidatorFound(ctx, cfg.API, validatorPubKey)
+	exist, err := isValidatorFound(ctx, cfg.StakingAPI, validatorPubKey)
 	if err != nil {
 		return err
 	}
@@ -403,11 +403,11 @@ func validateValidatorRedelegateFlags(ctx context.Context, cmd *cobra.Command, c
 		return errors.Wrap(err, "failed to decode hex-encoded pub key")
 	}
 
-	if cfg.API == "" {
+	if cfg.StakingAPI == "" {
 		return nil
 	}
 
-	existSrc, err := isValidatorFound(ctx, cfg.API, validatorSrcPubKey)
+	existSrc, err := isValidatorFound(ctx, cfg.StakingAPI, validatorSrcPubKey)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


client/cmd/flags.go
```diff
@@ -54,10 +63,11 @@ func bindInitFlags(flags *pflag.FlagSet, cfg *InitConfig) {
 }
 
 func bindValidatorBaseFlags(cmd *cobra.Command, cfg *baseConfig) {
-	cmd.Flags().StringVar(&cfg.RPC, "rpc", "https://storyrpc.io", "RPC URL to connect to the network")
+	cmd.Flags().StringVar(&cfg.RPC, "rpc", "https://mainnet.storyrpc.io", "RPC URL to connect to the network")
 	cmd.Flags().StringVar(&cfg.PrivateKey, "private-key", "", "Private key used for the transaction")
 	cmd.Flags().StringVar(&cfg.Explorer, "explorer", "https://storyscan.xyz", "URL of the blockchain explorer")
 	cmd.Flags().Int64Var(&cfg.ChainID, "chain-id", 1514, "Chain ID to use for the transaction")
+	cmd.Flags().StringVar(&cfg.API, "api", "", "URL of consensus client API server")
```
LeoHChen, line 70: is there a valid default value?

0xHansLee: I have set the default value to an empty string("") from two perspectives:

1. There is no API endpoint for mainnet provided by the foundation.
2. I wanted to maintain backward compatibility with the existing CLI usage. If the API is not set, the verification using the API is skipped.

Should we set the default value for this flag to the one of mainnet?

jdubpark (dismissed):
Can you see if you can move DecodeString logic to isValidatorFound to reduce duplication of check?

0xHansLee:
> Can you see if you can move DecodeString logic to isValidatorFound to reduce duplication of check?

I tried, but the public key is derived in two different ways—one from the input and the other from the private key. I also tried using duck typing (like implementing `GetPubKey()` for each config), but it requires some refactoring of our command configuration. I think it would be a good idea to refactor our command structure in the near future.

client/cmd/flags.go
```diff
@@ -363,3 +633,80 @@ func validateCommissionRate(ctx context.Context, cfg *createValidatorConfig) err
...
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, fmt.Sprintf("%s/staking/validators/%s", endpoint, valEVMAddr), nil)
+	if err != nil {
+		return stypes.Validator{}, errors.Wrap(err, "failed to create request for getting validator")
+	}
+
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return stypes.Validator{}, errors.Wrap(err, "failed to get validator")
+	}
```
gezerui0124, line 698: check `resp.Status == 200`

gezerui0124 (approved):

client/cmd/flags.go
limengformal: @gezerui0124 Is this the staking API?

Also, do we have a list of APIs we can provide to external validators? I suppose validators will be asking for it.

client/cmd/flags.go
```diff
@@ -211,7 +223,29 @@ func validateValidatorCreateFlags(ctx context.Context, cmd *cobra.Command, cfg *
 		return err
 	}
 
-	return validateCommissionRate(ctx, cfg)
+	if err := validateCommissionRate(ctx, cfg); err != nil {
+		return err
+	}
+
+	if cfg.API == "" {
```
limengformal, line 230: Should we output a warning that no API is provided so some validations (validator existence) are skipped? They run the risks that the validator may not exist and fund is lost.

Same for all the api checks below

client/cmd/flags.go
```diff
@@ -211,7 +223,29 @@ func validateValidatorCreateFlags(ctx context.Context, cmd *cobra.Command, cfg *
...
+		return nil
+	}
+
+	validatorPubKey, err := validatorKeyFileToCmpPubKey(cfg.ValidatorKeyFile)
```
limengformal, line 234: Should we put it above API check, so even API is not provided validator key formats can be validated?

Same for all validator key checks or conversions below

client/cmd/flags.go
jdubpark: Yes, this is the staking API.

We could modify to use either the staking API or local Cosmos SDK's staking module's REST API (which uses a different endpoint).

client/cmd/flags.go
jdubpark: Yes, let's do that.

jdubpark: As @edisonz0718 discussed, we will create a validator if it doesn't exist before staking, so funds won't be lost in the case that a user ignores the warning & stakes to a non-existent validator.

client/cmd/flags.go
jdubpark: Yes, sounds good.

jdubpark: changed in beb55c0

client/cmd/flags.go
jdubpark: changed in beb55c0

gezerui0124:
> Also, do we have a list of APIs we can provide to external validators? I suppose validators will be asking for it.

@limengformal we could expose some common story APIs of our rpc nodes to public, for now I think only `/staking/validators` and `/staking/validators/{validator_address}`.

And we could also extend it if needed.

client/cmd/flags.go
jdubpark: Should we rename this to `--staking-api` and update description to remove confusion with other APIs (e.g. RPC)?

client/cmd/flags.go
```diff
@@ -377,3 +647,83 @@ func validateCommissionRate(ctx context.Context, cfg *createValidatorConfig) err
...
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, fmt.Sprintf("%s/staking/validators/%s", endpoint, valEVMAddr), nil)
+	if err != nil {
+		return stypes.Validator{}, errors.Wrap(err, "failed to create request for getting validator")
+	}
+
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return stypes.Validator{}, errors.Wrap(err, "failed to get validator")
+	}
```
jdubpark, line 712: If the request fails (wrong API value), instead of failing, we could emit a warning about failure to validate and proceed to execute the requested action (e.g. update commission).

client/cmd/flags.go
0xHansLee: @jdubpark Sounds good. Will make change as you suggested.

client/cmd/flags.go
0xHansLee: User can skip this validation by not using api flag, so strict validation would be good even if the request fails with api flag. wdyt?

client/cmd/validator.go
```diff
@@ -151,7 +152,7 @@ type genPrivKeyJSONConfig struct {
 }
 
 func loadEnv() {
-	if err := godotenv.Load("../.env"); err != nil {
```
0xHansLee, line 154: This is rolled back. This was due to my local debug env (used `/client` as working dir), so `.env` under the root should be fine by default.

client/cmd/flags.go
```diff
@@ -377,3 +657,88 @@ func validateCommissionRate(ctx context.Context, cfg *createValidatorConfig) err
...
+		return stypes.ErrCommissionGTMaxChangeRate
+	}
+
+	return nil
+}
+
+type Validator struct {
+	OperatorAddress string            `json:"operator_address"`
+	Commission      stypes.Commission `json:"commission"`
+}
```
0xHansLee, line 692: Instead of using `stypes.Validator`, redefined a simplified Validator struct and used here. Since the `unbonding_height` is returned in string, but it is int64 in `stypes.Validator`.

client/cmd/flags.go
```diff
@@ -58,6 +67,7 @@ func bindValidatorBaseFlags(cmd *cobra.Command, cfg *baseConfig) {
 	cmd.Flags().StringVar(&cfg.RPC, "rpc", "https://storyrpc.io", "RPC URL to connect to the network")
 	cmd.Flags().StringVar(&cfg.Explorer, "explorer", "https://storyscan.xyz", "URL of the blockchain explorer")
 	cmd.Flags().Int64Var(&cfg.ChainID, "chain-id", 1514, "Chain ID to use for the transaction")
+	cmd.Flags().StringVar(&cfg.StakingAPI, "staking-api", "", "URL of Staking API server")
```
limengformal, line 70: Can we point to localhost to query?

limengformal: Also need a list of external staking APIs that others can use. Otherwise, who can use this option?

client/cmd/flags.go
edisonz0718: I think this is not staking api. This is story's API. @gezerui0124

client/cmd/flags.go
edisonz0718: I think we should let validators open the port and use localhost to connect.

client/cmd/flags.go
jdubpark: sounds good

client/cmd/flags.go
0xHansLee: Right, then how about `story-api`?

client/cmd/flags.go
0xHansLee: Should we use localhost for default value then?

limengformal (approved):

limengformal (approved):

PR #480, opened 2025-02-15
feat(ci): fix gha workflow to push docker images

AndyBoWu:
**CONTEXT**
The workflow to push images to ghcr has some hiccup. 

**PR NOTES**
1/ Update workflow title
2/ add input for workflow_dispatch trigger, handle the case if version is provisioned manually
3/ optimize docker caching strategy

issue: none

### DIFF
--- a/.github/workflows/ci-docker-hub.yml
+++ b/.github/workflows/ci-docker-hub.yml
@@ -1,14 +1,23 @@
-name: Build and Push to Docker Hub
+name: -> PUSH to GitHub Container Registry
 
 on:
   workflow_dispatch:
+    inputs:
+      version:
+        description: 'Tag version to use if triggered manually'
+        required: true
+        default: 'latest'
   push:
     tags:
       - '*'
 
 jobs:
   # Build and upload the geth binary
+  timestamp:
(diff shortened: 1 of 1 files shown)


0xHansLee (approved):

.github/workflows/ci-docker-hub.yml
```diff
@@ -40,13 +54,12 @@ jobs:
...
 
-      - name: Dockerize the geth and bootnode binary
+      - name: Dockerize the story-node binary
         env:
           DOCKER_BUILDKIT: 1
         run: |
-          docker buildx create --use
           docker buildx build \
-            --platform linux/amd64,linux/arm64,linux/arm64/v8 \
+            --platform linux/amd64,linux/arm64 \
```
limengformal, line 62: What was aem64/v8?

AndyBoWu: In multi-arch images, Docker often uses linux/arm64/v8 to explicitly indicate the ARMv8 instruction set, but linux/arm64 and linux/arm64/v8 will almost always resolve to the same 64-bit ARM platform.

so, linux/arm64/v8 is a bit redundant, or it may be introduced back then for legacy reason.

wo-o (approved):
LGTM

PR #482, opened 2025-02-18
feat(app): add with comet flag

0xHansLee:
Added with-comet flag, which is a flag for running an app with CometBFT. The default value is `true`.
This is to support KYVE's state sync, which uses external CometBFT process that feeds state-sync snapshots.

Also some changes in CLI flags for running node are made per KYVE team's request. With this changes, customized pruning is supported with `--pruning custom` flag.

## Added flags

- `with-comet`: Run abci app embedded in-process with CometBFT
- `address`: Address of proxy app
- `transport`: Specify abci transport (socket | grpc)
- `pruning-keep-recent`: Number of recent heights to keep on disk (ignored if pruning is not 'custom')
- `pruning-interval`: Height interval at which pruned heights are removed from disk (ignored if pruning is not 'custom'), this is not used by this command but kept for compatibility with the complete pruning options

## Updated

Some flags are renamed. These would lead to breaking changes in CLI usage for operation.

- rename flags of `--log-color`, `--log-format` and `--log-level` to `--log_color`, `--log_format` and `--log_level`
- rename flags of `--snapshot-interval` and `--snapshot-keep-recent` to `--state-sync.snapshot-interval` and `--state-sync.snapshot-keep-recent`

issue: #487


jdubpark (dismissed):

client/app/abci.go
```diff
@@ -27,7 +27,7 @@ func newABCIWrapper(
 	app abci.Application,
 	finalizeCallback postFinalizeCallback,
 	multiStoreProvider multiStoreProvider,
-) *abciWrapper {
+) abci.Application {
```
ramtinms, line 30: nit: maybe not using the interface for constructor

ramtinms (dismissed):

client/app/abci.go
0xHansLee: Good point. Changed in [004b83d](https://github.com/piplabs/story/pull/482/commits/004b83dbbff5cb3df9b25104f79af6f1a36559ae).

0xHansLee:
I will merge this PR after testing of KSYNC.

client/config/testdata/default.toml
```diff
@@ -62,6 +66,8 @@ evm-build-delay = "600ms"
 # more time for block building while ensuring faster consensus blocks.
 evm-build-optimistic = true
 
+with-comet = true
```
gezerui0124, line 69: `address` and `transport`

client/config/config.go
```diff
@@ -137,6 +149,7 @@ var (
 		ExternalAddress:    "",
 		Seeds:              "",
 		SeedMode:           false,
+		WithComet:          true,
```
gezerui0124, line 152: default value of `address` and `transport`

client/config/config.toml.tmpl
```diff
@@ -62,6 +66,9 @@ evm-build-delay = "{{ .EVMBuildDelay }}"
 # more time for block building while ensuring faster consensus blocks.
 evm-build-optimistic = {{ .EVMBuildOptimistic }}
 
+# WithComet defines whether to run abci app with CometBFT.
+with-comet = {{ .WithComet }}
```
gezerui0124, line 70: `address` and `transport`

gezerui0124 (approved):

client/config/testdata/default.toml
0xHansLee: It is intended as done in cosmos (not included in config, but only via flag). Since the `address` and `transport` are needed only when `with-comet` is false.

jdubpark (approved):

PR #484, opened 2025-02-22
feat(x/evmstaking): recovery for lost delegations

jdubpark:
more detail in the [forum post](https://forum.story.foundation/t/recovery-mechanism-for-lost-delegation-funds-due-to-user-mistakes/36879)

issue: none


client/x/evmstaking/keeper/deposit.go
```diff
@@ -96,8 +99,78 @@ func (k Keeper) ProcessDeposit(ctx context.Context, ev *bindings.IPTokenStakingD
...
+					sdk.NewAttribute(types.AttributeKeyValidatorCmpPubKey, hex.EncodeToString(ev.ValidatorCmpPubkey)),
+					sdk.NewAttribute(types.AttributeKeySenderAddress, ev.OperatorAddress.Hex()),
+					sdk.NewAttribute(types.AttributeKeyTxHash, hex.EncodeToString(ev.Raw.TxHash.Bytes())),
+					sdk.NewAttribute(types.AttributeKeyRefundCompletionTime, completionTime.String()),
+				),
+			})
+		}()
+
+		// push the refund to the unbonding queue
+		_, err = k.stakingKeeper.SetUnbondingDelegationEntry(
```
0xHansLee, line 154: It is not actual unbonding, so I think we need to use `stakingKeeper.InsertUBDQueue` to add it to the unbonding queue and it's enough ([ref](https://github.com/piplabs/cosmos-sdk/blob/40741b13972bbdfd51568a37e5a7f7d397985782/x/staking/keeper/delegation.go#L492)). `SetUnbondingDelegationEntry` is a function that stores the unbonding delegation using an index that increments by 1, without adding it to the unbonding queue.

client/x/evmstaking/keeper/deposit.go
```diff
@@ -96,8 +99,78 @@ func (k Keeper) ProcessDeposit(ctx context.Context, ev *bindings.IPTokenStakingD
...
+		if err != nil {
+			return errors.Wrap(err, "get refund fee")
+		}
+		refundFeeAmount := amountCoin.Amount.Mul(math.NewInt(int64(refundFeeBps))).Quo(math.NewInt(10_000))
+		refundAmount := amountCoin.Amount.Sub(refundFeeAmount)
+
+		// refund completion time is 1 day after the deposit
+		completionTime := sdkCtx.BlockTime().Add(time.Hour * 24)
+
+		defer func() {
```
0xHansLee, line 121: Can we combine this defer statement into above one?

It would be good to add new error codes, where the error codes change depending on whether a refund occurs. If a refund occurs, additional event attributes can be included and emitted. Since a refund is also a type of delegation failure case, the event type should be `EventTypeDelegateFailure`.

client/x/evmstaking/keeper/deposit_test.go
0xHansLee, line 1: It would be good to cover the case when a validator does not exist.

client/x/evmstaking/keeper/validator.go
```diff
@@ -158,8 +158,21 @@ func (k Keeper) ProcessCreateValidator(ctx context.Context, ev *bindings.IPToken
...
+			"val_story", validatorAddr,
+			"val_pubkey", validatorPubkey.String(),
+		)
+
+		if err = k.CreateDelegation(
+			cachedCtx, validatorAddr.String(), delegatorAddr.String(), amountCoin,
+			stypes.FlexiblePeriodDelegationID, // use flexible period delegation id
+			int32(0),                          // use flexible staking period
+		); err != nil {
+			return errors.Wrap(err, "convert validator creation to delegation as fallback")
```
0xHansLee, line 174: Also, it would be good to define a new sentinel error for this case.

0xHansLee (commented):
This changes will require hardfork and upgrade handler, right?

.gitignore
```diff
@@ -30,3 +30,4 @@ go.work
 go.work.sum
 
 story
+!client/proto/story
```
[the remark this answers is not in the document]
jdubpark: Yes, otherwise it excludes changes made to the proto files, which is located in `client/proto/story/{moduleName}`

client/x/evmstaking/types/params.go
```diff
@@ -70,3 +78,15 @@ func ValidateMinPartialWithdrawalAmount(v uint64) error {
 
 	return nil
 }
+
+func ValidateRefundFeeBps(v uint32) error {
+	if v == 0 {
+		return fmt.Errorf("refund fee bps must be positive: %d", v)
+	}
```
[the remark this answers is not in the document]
jdubpark: True, we might want to set the refund fee to zero bps at some point (so there is no refund fee), which would fail with this check.

.gitignore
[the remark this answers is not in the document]
0xHansLee: On my env, it is not excluded when I made changes in `client/proto/story/{moduleName}`.

Even without this, there should be no reason for the changes in the proto file to be excluded. Could this be an issue related to the local environment setting or configuration?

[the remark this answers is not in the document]
jdubpark: removed in 1c39a74

client/x/evmstaking/keeper/deposit.go
```diff
@@ -96,8 +99,78 @@ func (k Keeper) ProcessDeposit(ctx context.Context, ev *bindings.IPTokenStakingD
 	}
 
 	val, err := k.stakingKeeper.GetValidator(cachedCtx, validatorAddr)
+
+	//nolint:nestif // nested ifs error handling
 	if errors.Is(err, stypes.ErrNoValidatorFound) {
-		return errors.WrapErrWithCode(errors.ValidatorNotFound, err)
+		log.Debug(cachedCtx, "Validator not found, refunding deposit minus the refund fee",
```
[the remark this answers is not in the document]
jdubpark: changed in c3350bc

client/x/evmstaking/types/params.go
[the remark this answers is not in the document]
jdubpark: changed in 518cf08

client/x/evmstaking/keeper/deposit.go
```diff
@@ -96,8 +99,78 @@ func (k Keeper) ProcessDeposit(ctx context.Context, ev *bindings.IPTokenStakingD
...
+		// the min refund fee amount will be `refundFeeBps * minDelegationAmount (1024) / 10_000bps`
+		refundFeeBps, err := k.RefundFeeBps(cachedCtx)
+		if err != nil {
+			return errors.Wrap(err, "get refund fee")
+		}
+		refundFeeAmount := amountCoin.Amount.Mul(math.NewInt(int64(refundFeeBps))).Quo(math.NewInt(10_000))
+		refundAmount := amountCoin.Amount.Sub(refundFeeAmount)
+
+		// refund completion time is 1 day after the deposit
+		completionTime := sdkCtx.BlockTime().Add(time.Hour * 24)
```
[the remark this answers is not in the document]
jdubpark: changed in be50cc0

client/x/evmstaking/keeper/deposit.go
jdubpark: combined in d96d7db

client/x/evmstaking/keeper/validator.go
jdubpark: fixed in df9e800

client/x/evmstaking/keeper/deposit_test.go
jdubpark: added test in 2693e14

client/x/evmstaking/keeper/deposit.go
jdubpark: fixed in f027788

If I understand your request correctly, we need to add `InsertUBDQueue` (add to queue) after `SetUnbondingDelegationEntry` (create a ubd with unique ID).

This is following `Undelegate` fn which calls `SetUnbondingDelegationEntry` and then `InsertUBDQueue` ([ref](https://github.com/piplabs/cosmos-sdk/blob/40741b13972bbdfd51568a37e5a7f7d397985782/x/staking/keeper/delegation.go#L1270-L1280)).

0xHansLee: What if we do not call `SetUnbondingDelegationEntry`? IIUC, it should be fine. So I thought that we can just call `InsertUBDQueue`.

client/proto/story/evmstaking/v1/types/params.proto
```diff
@@ -19,4 +19,10 @@ message Params {
   string ubi_withdraw_address = 4 [
     (gogoproto.moretags) = "yaml:\"ubi_withdraw_address\""
   ];
+  uint32 refund_fee_bps = 5 [
+    (gogoproto.moretags) = "yaml:\"refund_fee_bps\""
+  ];
+  uint32 refund_period = 6 [
+    (gogoproto.moretags) = "yaml:\"refund_period\""
+  ];
```
0xHansLee, line 27: Why don't use second for the unit of `refund_period`? It would be consistent with the unit of unbonding period.

client/x/evmstaking/keeper/deposit.go
0xHansLee: Ah, I thought the unbonding queue was independent of unbonding delegation. But as you mentioned, it seems to be required since our EndBlocker process withdraws by dequeuing and completing the unbonding. SGTM. We should call both 👍

client/proto/story/evmstaking/v1/types/params.proto
jdubpark: changed in 04cae7d to use `time.Duration`, which is the unit used in `unbonding_time` in the staking module ([ref](https://github.com/piplabs/cosmos-sdk/blob/c7845d7a9d54010e33b31f0d4267dedc3e6a0f96/proto/cosmos/staking/v1beta1/staking.proto#L388-L390))

0xHansLee (approved):
LGTM

client/x/evmstaking/keeper/validator.go
```diff
@@ -158,8 +158,21 @@ func (k Keeper) ProcessCreateValidator(ctx context.Context, ev *bindings.IPToken
...
-		return errors.WrapErrWithCode(errors.ValidatorAlreadyExists, err)
+	//nolint:nestif // nested ifs error handling
+	if _, err := skeeperMsgServer.CreateValidator(cachedCtx, msg); errors.Is(err, stypes.ErrValidatorOwnerExists) || errors.Is(err, stypes.ErrValidatorPubKeyExists) {
+		// if validator exists, convert the request to a delegation
+		log.Debug(cachedCtx, "Validator already exists, converting stake from validator creation to delegation",
+			"val_story", validatorAddr,
+			"val_pubkey", validatorPubkey.String(),
+		)
+
+		if err = k.CreateDelegation(
```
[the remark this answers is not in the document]
jdubpark: changed in a4a6999 with test cases added in 635bbae

To confirm:
- if token type matches, convert to self-delegation
- otherwise, refund (same flow as delegating to non-existent validator)

jdubpark:
tests for creating existing validator (self-delegation if stake type matches, otherwise refund) added in 635bbae

client/x/evmstaking/keeper/validator.go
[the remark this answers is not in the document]
gezerui0124: LGTM

gezerui0124 (approved):

Issue #151, opened 2024-09-27
Validator key encryption


edisonz0718:
## Description and context
story client and cosmos clients in general use un-encrypted validator key files. This can be a security risk especially for bigger validators. We should provide better options for validators to safeguard their keys.

## Suggested solution
There are certain existing solutions in cosmos ecosystem. But some of them only support ed25519 keys. tmkms can be an initial option since it requires no changes to the story client code. But the remote signing may not be reliable especially in unreliable network and faulty hardware.

A better approach is to load the encrypted keys and a separate passphrase from a remote server during the node startup and use the passphrase to decrypt and store the keys in memory.

## Definition of done
Explore and find or design the right solution
Fully implement and test the solution

LeoHChen:
In this document, we also recommended plaintext private key in .env file, which should be avoided. Please research and add support of encrypted key support.

https://docs.story.foundation/docs/validator-operations#overview

PR #494, opened 2025-02-28
feat(cli): add to encrypt priv key

0xHansLee:
Support validator priv key encryption. 

### `init` command

Operators can encrypt the newly generated private key with `--encrypt-priv-key` flag. Without this flag, no encryption is done for private key as before. The encrypted file is stored under `story/config/enc_priv_key.json`. 

For all validator CLI, the private key is retrieved from the encrypted private key file by decrypting it if the `enc_priv_key.json` file exists. Otherwise, it is retrieved from `.env` file.

### `encrypt` command 

To support the existing operators (using not encrypted key), one command is also added to encrypt the existing private key in `.env` file: `./story key encrypt`. It will generate the encrypted private key file under `story/config/enc_priv_key.json`. 

### `show` command

Also, introduced a new command to decrypt and show the encrypted key: `./story key show-encrypted`. With `--show-private` flag, user can show the private key. 

issue: #151

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -91,6 +91,10 @@
       "path": "detect_secrets.filters.allowlist.is_line_allowlisted"
     },
     {
+      "path": "detect_secrets.filters.common.is_baseline_file",
+      "filename": ".secrets.baseline"
+    },
+    {
       "path": "detect_secrets.filters.common.is_ignored_due_to_verification_policies",
       "min_level": 2
     },
@@ -163,14 +167,14 @@
         "filename": "client/app/prompt.go",
         "hashed_secret": "87c3f13b0b9706003444e7bb666b658396894dde",
         "is_verified": false,
-        "line_number": 90
+        "line_number": 89
       },
       {
         "type": "Secret Keyword",
         "filename": "client/app/prompt.go",
         "hashed_secret": "5ffe533b830f08a0326348a9160afafc8ada44db",
         "is_verified": false,
-        "line_number": 94
+        "line_number": 93
       }
     ],
     "client/cmd/flags.go": [
@@ -958,5 +962,5 @@
       }
     ]
   },
-  "generated_at": "2025-02-28T08:18:38Z"
+  "generated_at": "2025-03-11T06:18:47Z"
 }
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -91,10 +91,6 @@
       "path": "detect_secrets.filters.allowlist.is_line_allowlisted"
     },
     {
-      "path": "detect_secrets.filters.common.is_baseline_file",
-      "filename": ".secrets.baseline"
-    },
-    {
       "path": "detect_secrets.filters.common.is_ignored_due_to_verification_policies",
       "min_level": 2
     },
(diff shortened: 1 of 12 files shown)


client/cmd/flags.go
```diff
@@ -26,9 +28,6 @@ import (
 	"github.com/piplabs/story/lib/k1util"
 	"github.com/piplabs/story/lib/netconf"
 	"github.com/piplabs/story/lib/tracer"
-
-	// Used for ABI embedding of the staking contract.
-	_ "embed"
```
0xHansLee, line 31: Not used

.secrets.baseline
LeoHChen, line 1: what's this file?

0xHansLee: It is generated by [`detect-secrets`](https://github.com/Yelp/detect-secrets) to detect leakage of secrets. 
This file lists the secrets detected in our codebase. There are no actual secrets in this PR, but `detect-secrets` detects some codes as secret, so added to those in this file.

jdubpark: Yes, updates are needed for false-positive secret keys to pass PR checks.

client/app/prompt.go
```diff
@@ -0,0 +1,111 @@
...
+		}
+	}
+
+	return password, nil
+}
+
+// ValidatePasswordInput validates a strong password input for new accounts,
+// including a min length.
+func ValidatePasswordInput(input string) error {
+	if len(input) < minPasswordLength {
```
jdubpark, line 106: We could do more weak criteria, e.g. at least one number.

jdubpark (approved):

client/app/prompt.go
0xHansLee: Right. This validation is the same one from Prysm. We could add more.

client/app/privkey.go
```diff
@@ -98,3 +114,87 @@ func exists(file string) bool {
...
 }
+
+// EncryptedKeyRepresentation defines an internal representation of encrypted validator key.
+type EncryptedKeyRepresentation struct {
+	Crypto  map[string]interface{} `json:"crypto"` //nolint:revive // This is from Prysm.
+	Version uint                   `json:"version"`
+	Name    string                 `json:"name"`
+}
+
+func EncryptAndStoreKey(key privval.FilePVKey, filePath string) error {
```
ramtinms, line 125: I suggest we add some unit tests for EncryptAndStoreKey and LoadEncryptedPrivKey functions

ramtinms (approved):

client/app/privkey.go
0xHansLee: Good suggestion!

I refactored some codes for test ([60d1b4c](https://github.com/piplabs/story/pull/494/commits/60d1b4c744371f031a8144682eaa52033265737d)) and add some test cases ([5b6711c](https://github.com/piplabs/story/pull/494/commits/5b6711cf988161b507b29c421ae1699093c0502e)).

ramtinms (approved):
Thanks, looks great

gezerui0124 (approved):

Issue #498, opened 2025-03-05
Versioning Smart Contract


0xHansLee:
## Description and context

We have no version of our contracts. Need to add versioning rule to our contracts: IPTokenStaking, UBIPool, and UngradeEntrypoint.

## Suggested solution
Following [`semver`](https://semver.org/) for versioning contract.

reference (Optimism): https://devdocs.optimism.io/contracts-bedrock/policies/versioning.html

## Definition of done
Our each contract has version

Ramarti:
Thanks! Reference link is broken @0xHansLee

Ramarti:
If we deterministically deploy the new implementation of the contracts, it would be akin to version control.
Agree it would be easier with more explicit semantic version.

0xHansLee:
@Ramarti Updated the link.

edisonz0718:
This is duplicate of #362

0xHansLee:
@edisonz0718, right, didn't notice that. Will close as duplicated.

PR #503, opened 2025-03-06
feat(contracts): script to deploy IPTokenStaking impl

Ramarti:
Script to deterministically deploy a new IPTokenStaking implementation with rounded down unstake amounts.

issue: none

### DIFF
--- a/contracts/script/upgrades/DeployNewIPTokenStaking_V1_0_1.s.sol
+++ b/contracts/script/upgrades/DeployNewIPTokenStaking_V1_0_1.s.sol
@@ -1,0 +1,42 @@
+// SPDX-License-Identifier: GPL-3.0-only
+pragma solidity 0.8.23;
+
+import { Script } from "forge-std/Script.sol";
+import { console2 } from "forge-std/console2.sol";
+
+import { IPTokenStaking } from "../../src/protocol/IPTokenStaking.sol";
+import { Predeploys } from "../../src/libraries/Predeploys.sol";
+import { Create3 } from "../../src/deploy/Create3.sol";
+
+/**
+ * @title DeployNewIPTokenStakingImpl
+ * @notice Deploys a new implementation of IPTokenStaking contract to be used for upgrading
+ * @dev This script only deploys the implementation contract, it does not perform the upgrade
+ */
+contract DeployNewIPTokenStaking_V1_0_1 is Script {
+    function run() external {
+        uint256 deployerPrivateKey = vm.envUint("DEPLOYER_PRIVATE_KEY");
+        address deployer = vm.addr(deployerPrivateKey);
+        vm.startBroadcast(deployerPrivateKey);
+
+        Create3 create3 = Create3(Predeploys.Create3);
+
+        // Generate creation code for IPTokenStaking
+        bytes memory creationCode = abi.encodePacked(
+            type(IPTokenStaking).creationCode,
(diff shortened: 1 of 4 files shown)


contracts/script/upgrades/DeployNewIPTokenStaking_V1_0_1.s.sol
LeoHChen, line 1: If we created a new implementation, I think we'd create a release of this version. @limengformal

LeoHChen: Where is the original deployment script?

limengformal: @LeoHChen The original deployment was through predeployment in geth genesis file. This is the [script](https://github.com/piplabs/story/blob/main/contracts/script/GenerateAlloc.s.sol) that generated the predeployment in alloc field.

limengformal: @LeoHChen Current issue is that staking contract is in the same repo of story client. 

Do you suggest we create a story release that only updates the staking contract? There are other unreleased changes that are already included in the current release. I would suggest to include it in the next story release when we fully test and release the unreleased story client changes.

contracts/script/upgrades/DeployNewIPTokenStaking_V1_0_1.s.sol
```diff
@@ -0,0 +1,42 @@
...
+import { IPTokenStaking } from "../../src/protocol/IPTokenStaking.sol";
+import { Predeploys } from "../../src/libraries/Predeploys.sol";
+import { Create3 } from "../../src/deploy/Create3.sol";
+
+/**
+ * @title DeployNewIPTokenStakingImpl
+ * @notice Deploys a new implementation of IPTokenStaking contract to be used for upgrading
+ * @dev This script only deploys the implementation contract, it does not perform the upgrade
+ */
+contract DeployNewIPTokenStaking_V1_0_1 is Script {
```
limengformal, line 16: Is this specific for the new implementation of IPTokenStaking contract? I think it is more of a general implementation deployment script for IPTokenStaking contract. Since if IPTokenStaking contract changes, the deployment changes as well.

I wouldn't point it to a specific version.

0xHansLee: At least, we need to mark commit hash for this deployment.

0xHansLee (approved):
Seeing that it is urgent to deploy this contract and upgrade, looks good to me. But as mentioned by @limengformal, we need versioning system for contracts. The related issue is https://github.com/piplabs/story/issues/498

contracts/script/upgrades/DeployNewIPTokenStaking_V1_0_1.s.sol
Ramarti: I think smart contract versioning should be separate from CL client.
We can use package.json within contracts to track that, maybe have distinct tag.
Then every contract has it's own versions (right now, tracked by virtue of different impl addresses)

limengformal:
> Seeing that it is urgent to deploy this contract and upgrade, looks good to me. But as mentioned by @limengformal, we need versioning system for contracts. The related issue is #498

Let's merge and modify later

limengformal (approved):

Issue #505, opened 2025-03-10
Support story-geth snap sync


0xHansLee:
## Description and context

When syncing blocks with state sync (CL) and snap sync (EL), The download of EL chain data is stuck at 99.99%. Because it is impossible to fetch state trie data from disk if it falls behind more than 128 blocks. So need to send FCU/newPayload with the latest block to update header chain.

## Suggested solution

Always request `forkChoiceUpdated` with the latest head while EVM syncing.

## Definition of done
Enable sync blocks with state sync (CL) and snap sync (EL).

PR #506, opened 2025-03-10
feat(x/evmengine): support exec engine snap sync

0xHansLee:
Previously, when the execution engine was syncing, it would keep retrying until evm syncing is done. However, evm syncing is stuck while downloading chain data because the requested head falls behind by more than 128 blocks, making it impossible to retrieve the missing data from other peers.

To support snap sync, `newPayloadV3` and `forkChoiceUpdatedV3` requests are now made with the latest head while the execution engine is syncing. During syncing, `prepareProposal`, `processProposal`, and `postFinalize` are skipped.

issue: #505

### DIFF
--- a/client/x/evmengine/keeper/keeper.go
+++ b/client/x/evmengine/keeper/keeper.go
@@ -28,6 +28,8 @@ import (
 	"github.com/piplabs/story/lib/k1util"
 )
 
+var ErrExecEngSyncing = errors.New("execution engine is syncing")
+
 type Keeper struct {
 	cdc             codec.BinaryCodec
 	storeService    store.KVStoreService
@@ -56,6 +58,8 @@ type Keeper struct {
 		Height    uint64
 		UpdatedAt time.Time
 	}
+
+	isExecEngSyncing bool
 }
 
 func NewKeeper(
@@ -126,6 +130,15 @@ func (k *Keeper) SetBuildOptimistic(b bool) {
 // SetValidatorAddress sets the validator address.
 func (k *Keeper) SetValidatorAddress(addr common.Address) {
 	k.validatorAddr = addr
+}
+
+// IsExecEngSyncing returns if execution engine is syncing or not.
+func (k *Keeper) IsExecEngSyncing() bool {
+	return k.isExecEngSyncing
+}
+
+func (k *Keeper) SetIsExecEngSyncing(isSyncing bool) {
+	k.isExecEngSyncing = isSyncing
 }
 
 // RegisterProposalService registers the proposal service on the provided router.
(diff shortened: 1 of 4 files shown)


jdubpark (approved):

client/x/evmengine/keeper/keeper.go
```diff
@@ -56,6 +58,8 @@ type Keeper struct {
 		Height    uint64
 		UpdatedAt time.Time
 	}
+
+	isExecEngSyncing bool
```
ramtinms, line 62: Just for learning the code base, this keeper is always on a single go routine ?

ramtinms (approved):
Nice work

stevemilk (approved):

client/x/evmengine/keeper/keeper.go
0xHansLee: Right!

gezerui0124 (approved):

client/x/evmengine/keeper/msg_server.go
```diff
@@ -124,9 +127,7 @@ func (s msgServer) ExecutionPayload(ctx context.Context, msg *types.MsgExecution
 
 			return false, nil // Retry
 		} else if isSyncing(fcr.PayloadStatus) {
-			log.Warn(ctx, "Processing finalized payload halted while evm syncing (will retry)", nil, "payload_height", payload.Number)
-
-			return false, nil // Retry
```
jdubpark, line 129: Does snap sync require `return false, nil // Retry` line to be removed here?

Asking since this changes the expected behavior. Originally, we expected `isSyncing == true` to retry the fetch logic. But removing `return false, nil` means `retryForever` will stop retrying when the node is syncing (or accepted) and continue to the next logic of delivering EL events to CL handlers.

I believe this will lead to non-determinism for a node if its execution client keeps returning `SYNCING` or `ACCEPTED` and eventually exceed the expected block time and progresses to the next block with non-updated execution client (which shouldn't happen in the original behavior, where the node should hang as it awaits syncing).

0xHansLee: Yes it is required for snap sync by requesting `forkChoiceUpdated` with the latest block.

The `forkChoiceUpdated` here is just to set safe and finalized block of EL chains. No fetched data is not required and used for finalized CL block. Only the data from finalized msg is used to process `FinalizeBlock`. That means, in CL perspective, it is fine to process the block with the finalized msg (just EL is syncing), not leading to non-determinism.

As I [mentioned](https://github.com/piplabs/trust-story/issues/43#issuecomment-2727804187) in the trust issue, the non-updated payload is not imported immediately. It is stashed for the update later by `forkChoiceUpdated`.

PR #514, opened 2025-03-13
feat(evmengine): use current finalized block events & remove prev events

jdubpark:
Remove prev EL block events in CL block. Instead, fetch & use finalized EL block's events after CL block is finalized.

issue: none


client/proto/story/evmengine/v1/types/tx.proto
```diff
@@ -19,7 +19,6 @@ message MsgExecutionPayload {
   option (cosmos.msg.v1.signer) = "authority";
   string            authority           = 1;
   bytes             execution_payload   = 2;
-  repeated EVMEvent prev_payload_events = 3;
```
0xHansLee, line 22: Please mark reserve or deprecated, not removing the field for future backward or forward compatibility.
It can cause unexpected issue in deserialization.
https://stackoverflow.com/questions/71782502/what-happened-if-you-delete-a-field-on-protobuf

jdubpark: fixed in 5b55016

PR #525, opened 2025-03-28
chore(config): set default with comet true

0xHansLee:
Set the default value of `--with-comet` flag for Aeneid and Mainnet.

issue: none

### DIFF
--- a/client/config/config.go
+++ b/client/config/config.go
@@ -112,6 +112,7 @@ var (
 		ExternalAddress:    "",
 		Seeds:              "",
 		SeedMode:           false,
+		WithComet:          true,
 	}
 	StoryConfig = Config{
 		HomeDir:            DefaultHomeDir(),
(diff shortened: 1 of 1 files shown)


limengformal (approved):

client/config/config.go
```diff
@@ -112,6 +112,7 @@ var (
 		ExternalAddress:    "",
 		Seeds:              "",
 		SeedMode:           false,
+		WithComet:          true,
```
limengformal, line 115: Do we need to set `with-tendermint` to true as well?

client/config/config.go
```diff
@@ -133,6 +134,7 @@ var (
...
```
limengformal, line 137: Do we need to set `with-tendermint` to true as well?

limengformal (approved):

client/config/config.go
0xHansLee: No need to set `with-tendermint`. This is an alias of `with-comet`, so if it is not specified, it will be not used.

jdubpark (approved):

PR #531, opened 2025-04-30
test(x/evmengine): add unit tests for evmengine module

0xHansLee:
Added unit tests for evmengine module. The coverage of functions in keeper is increased from 64.9% to 84%.

issue: none

### DIFF
--- a/client/x/evmengine/keeper/abci.go
+++ b/client/x/evmengine/keeper/abci.go
@@ -230,7 +230,7 @@ func (k *Keeper) PostFinalize(ctx sdk.Context) error {
 	maxWithdrawals, err := k.evmstakingKeeper.MaxWithdrawalPerBlock(ctx)
 	if err != nil {
 		log.Error(ctx, "Starting optimistic build failed; get max withdrawal", err, logAttr)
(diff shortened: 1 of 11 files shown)


client/x/evmengine/keeper/abci.go
```diff
@@ -230,7 +230,7 @@ func (k *Keeper) PostFinalize(ctx sdk.Context) error {
 	maxWithdrawals, err := k.evmstakingKeeper.MaxWithdrawalPerBlock(ctx)
 	if err != nil {
 		log.Error(ctx, "Starting optimistic build failed; get max withdrawal", err, logAttr)
-		return errors.Wrap(err, "get max withdrawal per block")
+		return nil
```
ramtinms, line 233: this looks like a behaviour change, is it expected?

0xHansLee: Good point. In conclusion, yes. Because when error occurs from `PostFinalize`, it is just printed and ignored as it is just for building payload optimistically.

ramtinms (approved):
Nice work

stevemilk (approved):

PR #558, opened 2025-05-30
chore(genesis): update genesis execution hash for local

0xHansLee:
Updated the execution block hash for local network

issue: none

### DIFF
--- a/lib/netconf/local/genesis.json
+++ b/lib/netconf/local/genesis.json
@@ -89,7 +89,7 @@
     },
     "evmengine": {
       "params": {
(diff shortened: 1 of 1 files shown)


0xHansLee:
Need to update the execution block hash according to https://github.com/piplabs/story-geth/pull/130

0xHansLee:
The local network failed to run because the commit that updated the genesis block was missing in story-geth. With [this](https://github.com/piplabs/story-geth/pull/130) fix, the local network is now working properly.

Issue #523, opened 2025-03-26
Recurring Validator signing missed at 300-Block Intervals


zeroN0de:
## Description and context
We have observed a recurring issue where multiple validators experience a burst of Missing votes exactly at 300-block intervals (e.g., heights 300, 600, 900, etc.). Through CometBFT debug logs, it appears that certain validators’ precommit votes are not reaching the next block proposer in time at these intervals, causing these Missing votes. Key points include:

- This does not seem limited to a specific validator; the same pattern occurs across different validators and proposers.
- Analysis of the Execution Layer (Geth) logs and transaction characteristics shows no significant differences for these blocks.
- Reviewing custom modules (e.g., block rewards) in the Story Consensus Client did not reveal a direct link to the issue.
- Adjusting certain P2P parameters (e.g., `p2p.flush_throttle_timeout`, `consensus.peer_gossip_sleep_duration`) did **not** reduce network delays, and the Missing occurrences have persisted.
- The issue persists for multiple validators, with some consistently missing votes at these 300-block intervals despite parameter tuning and peer changes.

## Steps to reproduce
1. Operate a Story/CometBFT-based network for a sufficient duration.
2. Monitor validator signatures around block heights that are multiples of 300 (e.g., 300, 600, 900).
3. Observe that, at these specific block heights, some validators’ precommit votes fail to arrive in time for the next proposer, resulting in Missing votes.

## Experienced behavior
- Missing votes spike precisely at 300-block intervals, whereas block signing proceeds normally at other heights.
- Precommit votes from certain validators arrive too late, leading to a new round being started before those votes are accounted for.

## Expected behavior
- The network should not experience concentrated Missing votes at regular intervals.
- All precommit votes should be transmitted and processed on time, regardless of the block height.

## Solution recommendation
- Could deeper investigation into CometBFT’s P2P or network layer help explain periodic traffic surges or latency around 300-block intervals?  
- Is there any module or process (e.g., pruning, snapshotting, or reward distribution) that might be triggering specifically at these intervals, causing a sudden increase in network load or delay?  
- Would exploring more robust peer connections (e.g., Mesh Peering) be beneficial if it is determined to be a purely network-layer issue?  
- Is continued monitoring of debug logs and further data collection feasible to pinpoint the root cause of these synchronized Missing occurrences?

jdubpark:
It could be pruning which I believe the default is 300 block interval.

zeroN0de:
[...]

PR #560, opened 2025-06-09
chore(config): change the default value of pruning interval

0xHansLee:
Change the default value of pruning interval to 10. 

issue: #523


stevemilk (approved):

ramtinms (approved):
Thanks for investigating and fixing the issue.

Issue #527, opened 2025-04-15
Use Protobuf marshalling for EVM payload


0xHansLee:
## Description and context

Currently, JSON marshalling and unmarshalling are used for EVM execution payloads, which unnecessarily increases data size, leading to network overhead and larger CL block sizes. Switching to protobuf marshalling/unmarshalling would help reduce the data size more efficiently.

## Suggested solution

As done in [Omni](https://github.com/omni-network/omni/pull/2826), use the protobuf for EVM execution payload. This would require a hardfork. 

## Definition of done

The EVM execution payload will be encoded and decoded using protobuf marshalling and unmarshalling, which will help reduce the size of CL blocks. As with the previous, unknown fields in the execution payload must be strictly disallowed when using protobuf.

PR #567, opened 2025-06-20
feat(evmengine): use protobuf to encode payload

0xHansLee:
This changes to use protobuf encoding for payload. Ported from https://github.com/omni-network/omni/pull/2826.

issue: #527

### DIFF
--- a/client/x/evmengine/keeper/keeper.go
+++ b/client/x/evmengine/keeper/keeper.go
@@ -26,6 +26,7 @@ import (
 	"github.com/piplabs/story/lib/errors"
 	"github.com/piplabs/story/lib/ethclient"
 	"github.com/piplabs/story/lib/k1util"
+	"github.com/piplabs/story/lib/netconf"
 )
 
 var ErrExecEngSyncing = errors.New("execution engine is syncing")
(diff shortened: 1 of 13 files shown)


client/x/evmengine/keeper/keeper.go
```diff
@@ -157,15 +158,38 @@ func (k *Keeper) parseAndVerifyProposedPayload(ctx context.Context, msg *types.M
...
+		if msg.ExecutionPayloadDeneb != nil {
+			return engine.ExecutableData{}, errors.New("proto payload not allowed")
+		}
+		// validate execution payload
+		if err := types.ValidateExecPayload(msg); err != nil {
+			return engine.ExecutableData{}, errors.Wrap(err, "validate execution payload")
+		}
+		if err := json.Unmarshal(msg.ExecutionPayload, &payload); err != nil {
+			return engine.ExecutableData{}, errors.Wrap(err, "unmarshal payload")
+		}
```
jdubpark, line 192: Are we using the versioning branch check to make this binary backward-compatible (for a single binary?) If v1.4.0 is a hardfork, could we remove the else part since we don't need to keep the old decoding method?

stevemilk (approved):
lgtm

PR #569, opened 2025-07-10
fix(evmstaking): handle residual reward by version

0xHansLee:
In Story v1.2.1, the handling of residual rewards was upgraded. As a result, when syncing from the genesis block using a client running v1.2.1 or later, a state difference may occur at a certain block, leading to an app hash mismatch error. To address this issue, the residual reward handling logic has been corrected to properly account for the block height — applying the appropriate logic before and after the v1.2.1 upgrade point.

issue: none

### DIFF
--- a/lib/netconf/upgrades.go
+++ b/lib/netconf/upgrades.go
@@ -1,0 +1,82 @@
+package netconf
+
+import (
+	"errors"
+)
+
+const (
+	Virgil   = "virgil"
+	Ovid     = "v1.2.0"
+	Polybius = "polybius"
+
+	V121 = "v1.2.1"
+)
+
+var (
+	ErrUnknownChainID = errors.New("unknown chain ID")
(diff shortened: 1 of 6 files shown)


gezerui0124 (approved):

ramtinms (approved):

lib/netconf/upgrades.go
```diff
@@ -0,0 +1,82 @@
...
+import (
+	"errors"
+)
+
+const (
+	Virgil   = "virgil"
+	Ovid     = "v1.2.0"
+	Polybius = "polybius"
+
+	V121 = "v1.2.1"
```
LeoHChen, line 12: Is there a release name?  why V121 here? The others have names.

0xHansLee: @limengformal is right.

Issue #589, opened 2025-09-04
Add ability to track staked balance of an address


Spablob:
## Description and context
Hey!

For context, we are preparing a technical proposal for the governance process. However, there is a technical requirement that is already present in the public Story constitution (https://www.story.foundation/constitution.pdf) that needs to be addressed which states:

<img width="1365" height="72" alt="Image" src="https://github.com/user-attachments/assets/2e8b3357-b97f-446a-8565-3ba352b64b63" />

This requirement means it is required to know how many tokens are staked for a given address on a given block. Then, use this information to integrate with Snapshot platform.

## Suggested solution
After review of how Snapshot implements their voting strategy's code - what is being looked for is possibly similar to how `getEthBalance()` works to obtain native token ($IP) balance of an address. Then what Snapshot does is using an archival node to get the native token balance via `getEthBalance()` at a certain block in the past (the snapshot block).

Brainstorming: to address the Story constitution requirement - a new `getStakedBalance()` which returns the staked balance of an address could be the solution. Then, using an archival node it would be possible to get that amount for an arbitrary block in the past.

This is the starting point for the brainstorm. What do you think?

## Definition of done
It is possible to know how much an address has as staked balance on a given block.

PR #592, opened 2025-09-13
feat(api): add endpoint to query staked token amount

0xHansLee:
Added APIs to fetch the staked token amount.
Also, implemented `CreateQueryContext` to allow querying data at a specific block height via the `X-Block-Height` request header.

issue: #589


jdubpark (approved):

client/server/staking.go
```diff
@@ -625,6 +627,115 @@ func (s *Server) GetValidatorsByDelegatorAddressValidatorAddress(r *http.Request
...
+		})
+	}
+
+	return QueryStakedTokenByDelegatorAddressResponse{
+		DelegationStakedToken: delStakedToken,
+		Pagination:            queryResp.Pagination,
+	}, nil
+}
+
+func (s *Server) GetTotalStakedTokenByDelegatorAddress(r *http.Request) (resp any, err error) {
```
Spablob, line 685: Noticed while testing that `GetTotalStakedTokenByDelegatorAddress` and `GetStakedTokenByDelegatorAddress` when called right after staking - do not recognize the newly staked immediately. 

Only after a few seconds, does the API gives the correct staked amount. Is this expected?

Brainstorming: Perhaps when querying past blocks the code could revert if not enough time or blocks have passed since the block being queried?

0xHansLee: Yes, it is expected. The CL block processing the staking related events from EL is far behind 1 block of EL, so should wait for about 2-3 seconds.

> Perhaps when querying past blocks the code could revert if not enough time or blocks have passed since the block being queried?

If there is no data to return, it will return an error. Even when the data at a specific height block (due to far behind of the latest block) is pruned.

PR #602, opened 2025-10-24
chore(upgrade): set terence upgrade

0xHansLee:
finalize story v1.4.0 stable. In addition, name the upgrade `Terence` and add story-localnet to upgrade histories.

issue: none

### DIFF
--- a/lib/netconf/upgrades.go
+++ b/lib/netconf/upgrades.go
@@ -47,7 +47,7 @@ var UpgradeHistories = map[string]UpgradeMap{
 		Ovid:     4477880,
 		V121:     5084300,
 		Polybius: 8270000,
-		Terence:  100000000,
+		Terence:  100000000, // TODO: need to set actual upgrade height for story mainnet
 	},
 }
 


limengformal (changes requested):
Let's not update version number yet. Let's test and then update so aeneid upgrade goes to v1.4.0

lib/buildinfo/buildinfo.go
```diff
@@ -13,10 +13,10 @@ import (
...
 
 const (
-	VersionMajor = 1          // Major version component of the current release
-	VersionMinor = 3          // Minor version component of the current release
-	VersionPatch = 2          // Patch version component of the current release
-	VersionMeta  = "unstable" // Version metadata to append to the version string
+	VersionMajor = 1        // Major version component of the current release
+	VersionMinor = 4        // Minor version component of the current release
+	VersionPatch = 0        // Patch version component of the current release
+	VersionMeta  = "stable" // Version metadata to append to the version string
```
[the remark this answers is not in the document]
0xHansLee: Got it

jdubpark (approved):

0xHansLee:
@limengformal The upgrade height is set for Aeneid in [f0bbdaa](https://github.com/piplabs/story/pull/602/commits/f0bbdaa7dddd41ef08484468a2fa553f9665c9d7).

lib/netconf/upgrades.go
```diff
@@ -25,26 +24,30 @@ type UpgradeMap map[string]int64
...
-		V140:     100000000,
+		Terence:  10886688,
 	},
 	StoryChainID: {
 		Virgil:   809988,
 		Ovid:     4477880,
 		V121:     5084300,
 		Polybius: 8270000,
-		V140:     100000000,
+		Terence:  100000000,
```
limengformal: Do we need to set it here? Is it something not reachable?

0xHansLee: Cannot start. Yes, this is for the compatibility just is case of using this binary for mainnet by setting not reachable height for the upgrade.

PR #606, opened 2025-11-12
feat(staking): add pausing to staking contract

Spablob:
Adds pausing capability to the staking contract

issue: none

### DIFF
--- a/client/app/upgrades.go
+++ b/client/app/upgrades.go
@@ -2,6 +2,7 @@ package app
 
 import (
 	"fmt"
+	"github.com/piplabs/story/client/app/upgrades/horace"
 
 	storetypes "cosmossdk.io/store/types"
 	upgradetypes "cosmossdk.io/x/upgrade/types"
@@ -26,6 +27,7 @@ var (
 		v_1_2_0.Upgrade,
 		polybius.Upgrade,
 		terence.Upgrade,
+		horace.Upgrade,
 	}
 	// Forks are for hard forks that breaks backward compatibility.
 	Forks = []upgrades.Fork{
@@ -33,6 +35,7 @@ var (
 		v_1_2_0.Fork,
 		polybius.Fork,
 		terence.Fork,
+		horace.Fork,
 	}
 )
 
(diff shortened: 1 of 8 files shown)


Ramarti (changes requested):
Please address comments and also add the deploy script for new version.
That way we can ensure new admin is the timelock (get admin value from `owner()` using old implementation)

contracts/src/upgrades/IPTokenStaking.sol
```diff
@@ -0,0 +1,547 @@
...
+        _setMinUnstakeAmount(args.minUnstakeAmount);
+        _setMinCommissionRate(args.minCommissionRate);
+        _setFee(args.fee);
+    }
+
+    /// @notice Initializes the V2 contract.
+    /// @dev Only callable once at proxy deployment.
+    /// @param pauser1 The address of the first pauser.
+    /// @param pauser2 The address of the second pauser.
+    function initializeV2(address pauser1, address pauser2) external reinitializer(2) {
```
0xHansLee: qq. can we call this initializer even though it's the case of proxy upgrade?

contracts/src/upgrades/IPTokenStaking.sol
```diff
@@ -0,0 +1,547 @@
...
+    /// @dev The caller must provide the compressed public key that matches the expected EVM address.
+    /// Use this method to make sure the caller is the owner of the validator.
+    /// @param validatorCmpPubkey 33 bytes compressed secp256k1 public key of validator.
+    /// @param moniker The moniker of the validator.
+    /// @param commissionRate The commission rate of the validator.
+    /// @param maxCommissionRate The maximum commission rate of the validator.
+    /// @param maxCommissionChangeRate The maximum commission change rate of the validator.
+    /// @param supportsUnlocked Whether the validator supports unlocked staking.
+    /// @param data Additional data for the validator.
+    function createValidator(
```
0xHansLee: We don't need to add the `whenNotPaused` modifier here?

Ramarti: I see your point @0xHansLee .
Our initial thought was to start with whatever stops a possible loss of funds from the users.
Revisiting this I think we could go with 2 options
- Every non admin methos is pausable but `unstake`
- Every non admin method is pausable

Always allow users to withdraw during an attack vs the attack may be exploiting the unstake logic so nothing should come out.

I think we can start with everything pausable and post in the Forum, see how community feels

contracts/src/upgrades/IPTokenStaking.sol
```diff
@@ -0,0 +1,547 @@
...
+    /// @notice Initializes the V2 contract.
+    /// @dev Only callable once at proxy deployment.
+    /// @param pauser1 The address of the first pauser.
+    /// @param pauser2 The address of the second pauser.
+    function initializeV2(address pauser1, address pauser2) external reinitializer(2) {
+        require(pauser1 != address(0), "IPTokenStaking: Invalid pauser1");
+        require(pauser2 != address(0), "IPTokenStaking: Invalid pauser2");
+        __AccessControl_init();
+        __Pausable_init();
+        _grantRole(DEFAULT_ADMIN_ROLE, 0x6c7FA8DF1B8Dc29a7481Bb65ad590D2D16787a82);
```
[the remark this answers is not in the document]
Spablob: added comment in https://github.com/piplabs/story/pull/606/commits/48d61107730424c41e2c935c65cafb512d5a30cc

PR #597, opened 2025-10-17
chore(deps): upgrade go version and deps

0xHansLee:
Upgrade go version and deps.
This PR covers the below PRs created by dependabot:

- https://github.com/piplabs/story/pull/559
- https://github.com/piplabs/story/pull/574
- https://github.com/piplabs/story/pull/583
- https://github.com/piplabs/story/pull/586
- https://github.com/piplabs/story/pull/587

Also, it covers some high severity security alerts from dependabot:

- https://github.com/piplabs/story/security/dependabot/29

And it resolves the issue #556 

issue: none

PR #655, opened 2026-02-03
fix(upgrade): skip to update and check 0 type of period in virgil handler

0xHansLee:
At the Virgil upgrade, we changed the parameters for period types 1, 2, and 3, excluding type 0.
To properly handle period type 0 at the Virgil upgrade block, the handler should skip the other period types instead of returning an error.

This modification was introduced in #597 to fix golang lint.

issue: none


stevemilk:
Nodes that have passed Virgil are not impacted, while nodes run full sync from genesis will be impacted by this issue, correct?

jdubpark (approved):

0xHansLee:
Hey @stevemilk and @jdubpark, this [commit](https://github.com/piplabs/story/pull/655/commits/fe445dbd69d2497b4c87432afbe5bc0a682b5f19) is adding a unit test for virgil upgrade handler to pass the codecov check. Can you check this?

stevemilk (approved):

PR #672, opened 2026-02-12
fix(dkg): remove deals from session state

stevemilk:
issue: #673
The CL now filters deals by recipient index and forwards them to the kernel for storage. Therefore, there is no need to maintain deals on the CL side anymore.

### DIFF
--- a/client/x/dkg/keeper/dkg_svc_dealing.go
+++ b/client/x/dkg/keeper/dkg_svc_dealing.go
@@ -2,12 +2,13 @@ package keeper
 
 import (
 	"context"
-	"cosmossdk.io/collections"
 	"encoding/hex"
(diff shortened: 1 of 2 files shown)


stevemilk:
@0xHansLee Should the kernel also filter deals? It can reject misrouted ones anyway, but filtering first would save unnecessary work.

0xHansLee:
> @0xHansLee Should the kernel also filter deals? It can reject misrouted ones anyway, but filtering first would save unnecessary work.

Yeah, it's for just in case. The kernel skips misrouted deals or responses.

0xHansLee:
Please check the failed gha. The issue footer is missed.

0xHansLee (approved):

PR #676, opened 2026-02-19
feat(dkg): move finalization signature verification from contract to cl

0xHansLee:
## Summary

Move DKG finalization signature verification from the DKG contract to the consensus layer (`dkg_handler.go`) so that TEE signature validation is enforced at the CL before state changes.

- Add `verifyFinalizationSignature` in `dkg_handler.go` that reproduces the TEE-signed message hash (codeCommitment + round + participantsRoot + globalPubKey + publicCoeffs), recovers the ECDSA signer via eth_sign prefix, and checks it matches address(keccak256(commPubKey))
- Integrate signature verification into `Finalized()` after registration lookup and before vote counting
- Remove `_verifyFinalizationSignature` and its require call from `DKG.sol`, drop unused ECDSA/MessageHashUtils imports
- Remove `testFinalizeDKG_RevertIfInvalidSignature` from `dkg.t.sol` and clean up unused fixture variables
- Add comprehensive tests: valid/tampered signatures, wrong signer, commPubKey length, empty coefficients, V=28 recovery, double finalization, threshold global key setting

issue: none

### DIFF
--- a/client/x/dkg/keeper/dkg_handler.go
+++ b/client/x/dkg/keeper/dkg_handler.go
@@ -3,6 +3,7 @@ package keeper
 import (
 	"bytes"
 	"context"
+	"encoding/binary"
 	"encoding/hex"
 	"fmt"
 	"slices"
@@ -11,7 +12,6 @@ import (
 	"github.com/ethereum/go-ethereum/crypto"
 
 	"github.com/ethereum/go-ethereum/common"
-
 	"github.com/piplabs/story/client/x/dkg/types"
 	"github.com/piplabs/story/lib/errors"
 	"github.com/piplabs/story/lib/log"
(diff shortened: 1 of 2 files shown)


client/x/dkg/keeper/dkg_handler.go
```diff
@@ -114,6 +115,21 @@ func (k *Keeper) Finalized(ctx context.Context, round uint32, msgSender common.A
...
+	if err != nil {
+		return errors.Wrap(err, "failed to get DKG registration for signature verification")
+	}
+
+	// Prevent double-finalization: reject if this validator already finalized for this round
+	if reg.Status == types.DKGRegStatusFinalized {
+		return errors.New("validator has already finalized for this round")
+	}
+
+	if err := verifyFinalizationSignature(reg.CommPubKey, round, codeCommitment, participantsRoot, globalPubKey, publicCoeffs, signature); err != nil {
```
stevemilk: qq: Is it true that one node will fail consensus if it cannot verify the signature correctly since dkgkeeper's kvstore is now part of the consensus?

0xHansLee: Yes, this is deterministic. If one fails to do this, every nodes fail, so rejecting handling this.

stevemilk (approved):

PR #678, opened 2026-02-25
feat(dkg): track pubkeyshare

stevemilk:
The kernel returns `pubKeyShare` in FinalizeDkg call and CL stores it when updating DKGRegistrationStatus. 
This is preparation for partial decryption.

issue: none


client/x/dkg/keeper/dkg_handler.go
```diff
@@ -127,7 +128,7 @@ func (k *Keeper) Finalized(ctx context.Context, round uint32, msgSender common.A
 		}
 	}
 
-	if err := k.updateDKGRegistrationStatus(ctx, codeCommitment, round, msgSender, types.DKGRegStatusFinalized); err != nil {
+	if err := k.updateDKGRegistrationStatus(ctx, codeCommitment, round, msgSender, pubKeyShare, types.DKGRegStatusFinalized); err != nil {
```
0xHansLee: This is to update the status of dkg registration. Can you use `k.setDKGRegistration` instead?

stevemilk: i prefer renaming this function to avoid calling `getDKGRegistration` twice, how do you think

0xHansLee: Previously, there were several steps, so this function was used for that. Now, it is used is finalizing only, so I think it's good. LGTM

client/x/dkg/keeper/dkg_handler.go
```diff
@@ -128,7 +128,7 @@ func (k *Keeper) Finalized(ctx context.Context, round uint32, msgSender common.A
 		}
 	}
 
-	if err := k.updateDKGRegistrationStatus(ctx, codeCommitment, round, msgSender, pubKeyShare, types.DKGRegStatusFinalized); err != nil {
+	if err := k.finalizeDKGRegistration(ctx, codeCommitment, round, msgSender, pubKeyShare, types.DKGRegStatusFinalized); err != nil {
```
0xHansLee: Then, we don't need to pass the `types.DKGRegStatusFinalized` as an arg here.

0xHansLee (approved):

PR #681, opened 2026-02-25
feat(cdr): add cdr draft smart contracts

Spablob:
Adds cdr draft smart contracts

issue: none

PR #682, opened 2026-02-27
feat(dkg): collect partial decryptions

stevemilk:
Changes:
- Verify partial decryption response by decoding signature and comparing pubKeyShare
- Store the partial decryptions in the keeper’s KVStore and make them available for the requester to fetch.
- Use GlobalPubKey instead of DkgPubKey during partial decryption
- Add timeout mechanism for decryption requests

issue: none


contracts/src/interfaces/IDKG.sol
```diff
@@ -162,9 +198,42 @@ interface IDKG {
...
+    /// @param ciphertext The ciphertext to decrypt
+    /// @param label The request label
+    function requestThresholdDecryption(
+        uint32 round,
+        bytes32 codeCommitment,
+        bytes calldata requesterPubKey,
+        bytes calldata ciphertext,
+        bytes calldata label
+    ) external;
+
```
0xHansLee: These would be in CDR contract as you can see #681. `read` would be the request for threshold decryption and `submitEncryptedPartialDecryption` would be for the submission. After merging the PR, we can update the dkg module.

stevemilk: sounds good.

client/x/dkg/keeper/abci.go
```diff
@@ -52,6 +52,14 @@ func (k *Keeper) BeginBlocker(ctx context.Context) error {
 			// round = DKGStageRegistration if either
 			// 1. it's the initial (first) round, OR
 			// 2. the active stage of the previous round has ended, so DKG needs to reshare deals
+			//
+			// shouldTransitionStage only returns DKGStageRegistration when the previous stage
+			// was DKGStageActive, so sweep the decrypt request registry for the ending round.
+			var prevCC [32]byte
+			copy(prevCC[:], latestRound.CodeCommitment)
+			if err := k.sweepDecryptRequestRegistry(ctx, prevCC, latestRound.Round); err != nil {
```
0xHansLee: If a partial decryption request occurs and a new round begins within the submission timeout (200 blocks in this PR), the DKG members from the previous active round should continue generating partial decryptions until the new round becomes active. In other words, when a new round starts, it does not seem necessary to reset the decryptRequestRegistry. Wdyt?

client/x/dkg/keeper/dkg_decrypt_registry.go
```diff
@@ -0,0 +1,82 @@
...
+	key := decryptRequestRegistryKey(codeCommitment, round, label)
+	if err := k.DecryptRequestRegistry.Remove(ctx, key); err != nil {
+		return errors.Wrap(err, "delete decrypt request registry entry")
+	}
+	return nil
+}
+
+// sweepDecryptRequestRegistry deletes all registry entries for the given
+// (codeCommitment, round) pair. Should be called when a DKG round ends.
+func (k *Keeper) sweepDecryptRequestRegistry(ctx context.Context, codeCommitment [32]byte, round uint32) error {
```
0xHansLee: Instead of scanning the entire map and filtering manually, `Range.Prefix()` leverages the IAVL tree's sorted structure to seek directly to matching keys — reducing the cost from O(n) over all entries to O(k log n) where k is only the number of matching entries. The below is the re-written code using the ranger. 

```go
func (k *Keeper) sweepDecryptRequestRegistry(ctx context.Context, codeCommitment [32]byte, round uint32) error {
	prefix := fmt.Sprintf("%s_%d_", hex.EncodeToString(codeCommitment[:]), round)

	rng := new(collections.Range[string]).Prefix(prefix)

	iter, err := k.DecryptRequestRegistry.Iterate(ctx, rng)
	if err != nil {
		return errors.Wrap(err, "iterate decrypt request registry")
	}
	defer iter.Close()

	keys, err := iter.Keys()
	if err != nil {
		return errors.Wrap(err, "collect decrypt request registry keys")
	}

	for _, key := range keys {
		if err := k.DecryptRequestRegistry.Remove(ctx, key); err != nil {
			return errors.Wrap(err, "remove decrypt request registry entry")
		}
	}

	return nil
}
```

client/x/dkg/keeper/abci.go
stevemilk: You right. I overlooked that history dkg network still exists and requests in previous dkg round should be handled. Then I think we can start a new routine to clean the requestRegistry rather than sweep at the start of each round.

0xHansLee: What does it mean of starting a new routine at the beginning of each round? Can you elaborate it?

stevemilk: A cronjob that cleans expired decrypt requests.

stevemilk: Cronjob might not work. It breaks consensus. Will change it

client/x/dkg/keeper/keeper.go
```diff
@@ -48,13 +49,16 @@ type Keeper struct {
 
 	Schema             collections.Schema
 	ParamsStore        collections.Item[types.Params]
-	DKGNetworks        collections.Map[string, types.DKGNetwork]        // key: round
-	LatestDKGNetwork   collections.Item[string]                         // stores key of latest DKG network
+	DKGNetworks        collections.Map[string, types.DKGNetwork]        // key: codeCommitment_round
```
0xHansLee: From this PR, we don't use code commitment on CL anymore. All verifications related to code commitment will be managed on DKG contract, and the code commitment is managed by governance (owner of the contract).

client/x/evmengine/keeper/dkg.go
```diff
@@ -415,3 +415,152 @@ func (k *Keeper) ProcessDKGUpgradeCancelled(ctx context.Context, ethlog *ethtype
 
 	return nil
 }
+
+// func (k *Keeper) ProcessDKGDealComplaintsSubmitted(ctx context.Context, ethlog *ethtypes.Log) (err error) {
```
0xHansLee: We don't need this commented codes

contracts/src/interfaces/ICDR.sol
```diff
@@ -41,15 +41,43 @@ interface ICDR {
...
-    /// @param encryptedData The encrypted data
-    /// @param recipientPublicKey The uncompressed public key of the recipient
-    event VaultRead(uint32 uuid, bytes encryptedData, bytes recipientPublicKey);
+    /// @param requester The address requesting the read (msg.sender)
+    /// @param round The DKG round number for threshold decryption
+    /// @param codeCommitment The DKG code commitment identifying the committee
+    /// @param ciphertext The encrypted data (ciphertext)
+    /// @param requesterPubKey The public key of the requester
+    /// @param label The label identifying the decrypt request
+    event VaultRead(
```
0xHansLee, line 47: Why do you made the changes in the interface?

client/x/dkg/keeper/dkg_decrypt_registry.go
```diff
@@ -0,0 +1,91 @@
...
+	"cosmossdk.io/collections"
+
+	"github.com/piplabs/story/client/x/dkg/types"
+	"github.com/piplabs/story/lib/errors"
+)
+
+// decryptRequestRegistryKey builds the key for the DecryptRequestRegistry map:
+// hex(codeCommitment)_round_hex(sha256(label))
+func decryptRequestRegistryKey(codeCommitment [32]byte, round uint32, label []byte) string {
+	labelHash := sha256.Sum256(label)
```
0xHansLee: The label would be [32]byte of data. Should we hash this?

client/x/dkg/keeper/dkg_decrypt_registry.go
```diff
@@ -0,0 +1,91 @@
...
+	return fmt.Sprintf("%s_%d_%s", hex.EncodeToString(codeCommitment[:]), round, hex.EncodeToString(labelHash[:]))
```
0xHansLee: No code commitment would be required here I think. Also, round. I think the label is enough to specify the decrypt request. wdyt?

client/x/dkg/keeper/abci.go
```diff
@@ -58,6 +58,14 @@ func (k *Keeper) BeginBlocker(ctx context.Context) error {
 		return k.InitiateDKGRound(ctx, true)
 	}
 
+	// Prune timed-out decrypt request registry entries every N blocks.
+	// All nodes evaluate this condition identically, preserving consensus.
+	if currentHeight%types.DecryptRequestRegistryCleanupInterval == 0 {
```
0xHansLee, line 64: Should this be executed before checking if there is any upgrade pending? When upgrade is activated, it will initiate a new round of DKG and return before reaching to this check.

client/x/dkg/keeper/dkg_partial_decryption.go
```diff
@@ -0,0 +1,60 @@
...
+	CodeCommitment   []byte `json:"code_commitment"`
+	Pid              uint32 `json:"pid"`
+	EncryptedPartial []byte `json:"encrypted_partial"`
+	EphemeralPubKey  []byte `json:"ephemeral_pub_key"`
+	PubShare         []byte `json:"pub_share"`
+	Label            []byte `json:"label"`
+}
+
+func dkgPartialDecryptKey(codeCommitment [32]byte, round uint32, label []byte) string {
+	labelHash := sha256.Sum256(label)
```
0xHansLee: Same here. The label would be enforced by [32]byte.

client/x/dkg/keeper/dkg_decrypt_registry.go
stevemilk: how are labels generated?

0xHansLee: label is the uuid from CDR contract.

https://github.com/piplabs/story/blob/f3451a572b79eb9bcbb76e3c773becf2f85b4f92/contracts/src/protocol/CDR.sol#L86

contracts/src/interfaces/ICDR.sol
stevemilk: it's from story-kernel code, i basically copy the structure of `PartialDecryptTDH2Request`.

client/x/dkg/keeper/dkg_decrypt_registry.go
stevemilk: if so i think no need

client/x/dkg/keeper/dkg_decrypt_registry.go
stevemilk: if many `read` on the same vault, maybe label+salt can represent one request?

client/x/dkg/keeper/abci.go
stevemilk: yes we should not call this before decryptRequestRegistry is added in this upcoming upgrade.

contracts/src/interfaces/ICDR.sol
0xHansLee: I know. Can you check the story-kernel side if we need these fields even now?

client/x/dkg/keeper/dkg_decrypt_registry.go
0xHansLee: Maybe label + recipientPublicKey or recipient address?

contracts/src/interfaces/ICDR.sol
stevemilk: Checked, I think we can remove `CodeCommitment` and `SealedShareId`

client/x/dkg/keeper/abci.go
0xHansLee: I see. That means, if the upgrade is activated at for example 1000, this pruning will be skipped at the upgrade height and will be done at 2000, right? Is this expected behavior?

contracts/src/interfaces/ICDR.sol
0xHansLee: Sounds good. We can use the latest active round for Round here then. Do we need to keep the `EphemeralPubKey` in submission of partial?

client/x/dkg/keeper/dkg_decrypt_registry.go
stevemilk: one user can read one vault multiple times. The salt needs to be a random value or just count from 1.

client/x/dkg/keeper/abci.go
stevemilk: Right

contracts/src/interfaces/ICDR.sol
stevemilk: Yes users need it to parse the decryptPartials

client/x/dkg/keeper/dkg_decrypt_registry.go
0xHansLee: Then, the user can use different recipient public key? The eligibility check can be done if the account has IP License token or not.

stevemilk: Is it necessary to do the eligibility check? I thought users can specify any public key though normally it should be the key of the msg.sender. Then they can use the corresponding account to fetch the partials.

0xHansLee: > I thought users can specify any public key though normally it should be the key of the msg.sender.

Right. It is not enforced that the msg.sender is from the recipient public key. The eligibility check is an example of read condition. It would be handled in `readConditionAddr` contract for each vault.

ramtinms (approved):

client/x/dkg/keeper/dkg_decrypt_registry.go
```diff
@@ -0,0 +1,89 @@
...
+	"cosmossdk.io/collections"
+
+	"github.com/piplabs/story/client/x/dkg/types"
+	"github.com/piplabs/story/lib/errors"
+)
+
+// decryptRequestRegistryKey builds the key for the DecryptRequestRegistry map:
+// hex(codeCommitment)_round_hex(sha256(label))
+func decryptRequestRegistryKey(round uint32, label []byte) string {
+	return fmt.Sprintf("%d_%s", round, hex.EncodeToString(label))
```
0xHansLee: Why don't we remove the round here as well? I think it's enough to use label here. 

The reason why I'm suggesting is there can be the case the the round when read request and the partial submitted can be different.

contracts/src/protocol/CDR.sol
```diff
@@ -146,12 +146,16 @@ contract CDR is ICDR, Ownable2StepUpgradeable, ReentrancyGuardUpgradeable, Pausa
...
-    /// @param recipientPublicKey The public key of the recipient
+    /// @param requesterPubKey The public key of the requester
+    /// @param label The label identifying the decrypt request
     function read(
         uint32 uuid,
+        uint32 round,
         bytes memory accessAuxData,
-        bytes calldata recipientPublicKey
+        bytes calldata requesterPubKey,
+        bytes calldata label
```
0xHansLee: We don't need this. `uuid` would be used as `label` in threshold decryption.

contracts/src/protocol/CDR.sol
```diff
@@ -174,26 +178,36 @@ contract CDR is ICDR, Ownable2StepUpgradeable, ReentrancyGuardUpgradeable, Pausa
...
+    /// @param signature The signature over the partial decryption payload
     function submitEncryptedPartialDecryption(
-        address enclaveID,
+        uint32 round,
+        bytes32 codeCommitment,
+        uint32 pid,
         bytes calldata encryptedPartial,
+        bytes calldata ephemeralPubKey,
+        bytes calldata pubShare,
+        bytes calldata label,
```
0xHansLee: Please rename this to `uuid` in CDR contract.

contracts/src/protocol/CDR.sol
```diff
@@ -174,26 +178,36 @@ contract CDR is ICDR, Ownable2StepUpgradeable, ReentrancyGuardUpgradeable, Pausa
...
     /// @param encryptedPartial The encrypted partial decryption
-    /// @param signature The signature of the encrypted partial decryption
+    /// @param ephemeralPubKey The ephemeral public key used for encryption
+    /// @param pubShare The validator's public key share
+    /// @param label The label identifying the decrypt request (vault UUID bytes)
+    /// @param signature The signature over the partial decryption payload
     function submitEncryptedPartialDecryption(
-        address enclaveID,
+        uint32 round,
+        bytes32 codeCommitment,
```
0xHansLee: I don't think code commitment is needed here.

contracts/src/protocol/CDR.sol
```diff
@@ -146,12 +146,16 @@ contract CDR is ICDR, Ownable2StepUpgradeable, ReentrancyGuardUpgradeable, Pausa
...
     /// @notice Reads data from a vault
     /// @param uuid The UUID of the vault
+    /// @param round The DKG round number for threshold decryption
     /// @param accessAuxData The auxiliary access data for reading
-    /// @param recipientPublicKey The public key of the recipient
+    /// @param requesterPubKey The public key of the requester
+    /// @param label The label identifying the decrypt request
     function read(
         uint32 uuid,
+        uint32 round,
```
0xHansLee: I think user don't need to put round in `read` request. It should be determined when processing the request.

client/x/evmengine/keeper/cdr.go
```diff
@@ -41,16 +48,6 @@ func (k *Keeper) ProcessCDRVaultRead(ctx context.Context, ethlog *ethtypes.Log)
 		return errors.Wrap(err, "parse VaultRead log")
 	}
 
-	latestActive, err := k.dkgKeeper.GetLatestActiveRound(ctx)
```
0xHansLee, line 46: And if there is no round in the event, we can use the latest active round here.

client/x/dkg/keeper/keeper.go
```diff
@@ -49,12 +50,15 @@ type Keeper struct {
 	Schema             collections.Schema
 	ParamsStore        collections.Item[types.Params]
 	DKGNetworks        collections.Map[string, types.DKGNetwork]        // key: round
-	LatestDKGNetwork   collections.Item[string]                         // stores key of latest DKG network
+	LatestDKGNetwork   collections.Item[string]                         // stores codeCommitment key of latest DKG network
```
0xHansLee: Please remove codeCommitment in comment in this file. The code commitment will not be used in CL anymore.

client/x/dkg/keeper/dkg_partial_decryption.go
```diff
@@ -0,0 +1,58 @@
...
+	"fmt"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/piplabs/story/lib/errors"
+)
+
+type partialDecryptionSubmission struct {
+	Validator        string `json:"validator"`
+	Round            uint32 `json:"round"`
+	CodeCommitment   []byte `json:"code_commitment"`
```
0xHansLee: I think we don't need code commitment in partial decryption.

client/x/dkg/keeper/dkg_decrypt_registry.go
stevemilk: replaced to requesterPubkKey+label

contracts/src/interfaces/ICDR.sol
```diff
@@ -91,16 +113,30 @@ interface ICDR {
...
-        address enclaveID,
+        uint32 round,
+        uint32 pid,
         bytes calldata encryptedPartial,
+        bytes calldata ephemeralPubKey,
+        bytes calldata pubShare,
+        bytes calldata requesterPubKey,
+        uint32 uuid,
         bytes calldata signature
     ) external payable;
```
0xHansLee, line 141: Do we need `round`, `requesterPubKey` here?

stevemilk: `round` is for fetching registrations, `requesterPubKey` is used in decryptRequestRegistryKey mentioned above

0xHansLee: Can we fetch registrations of latest active round? not using the `round` here?

stevemilk: I think we cannot. Registrations may change each round. One validator can also change its commPubkey in different rounds.

0xHansLee: Yeah, it's for verifying the signature by fetching the comm pub key, right? I think only the DKG members of recent active round can submit the partials. If the round ends after the request is submitted and the next round is activated, the submission from previous round should be rejected. Does this make sense?

client/x/dkg/keeper/dkg_handler.go
0xHansLee, line 454: Can you make this happen only when DKG service is enabled?

0xHansLee (approved):

Issue #692, opened 2026-03-07
Implement the upgrade handler to activate the DKG module


0xHansLee:
## Description and context

Implement the upgrade handler to activate the DKG module

- [ ] Implement upgrade handler: `RunMigrations` (auto-InitGenesis for DKG) + set default DKG params + enable vote extensions at `upgradeHeight + 1`
- [ ] Gate all DKG functionality behind `netconf.IsV200` height checks:
  - `BeginBlocker`: skip DKG round initiation before activation
  - `ExtendVote` / `VerifyVoteExtension`: return no-op before activation
  - `PrepareVotes`: return `nil` (not empty `MsgAddDkgVote`) before `v200Height + 2` to maintain proposal compatibility
  - `ProcessProposal`: conditionally expect `MsgAddDkgVote` only after `v200Height + 1`
  - `PrepareProposal`: conditionally include `MsgAddDkgVote` only when `PrepareVotes` returns non-nil
  - UBI reward distribution: skip DKG committee distribution before activation
- [ ] Modify `UpgradeStoreLoader` to pre-add stores from all upgrades (past and future) so the new binary can start at any height without "store version mismatch" errors
- [ ] Evaluate whether rolling upgrade (mixed old/new binary coexistence before upgrade height) is feasible

### Rolling upgrade feasibility analysis

After thorough investigation of Cosmos SDK's `rootmulti.Store` internals, I concluded that **rolling upgrade is not possible** when adding a new module KVStore:

1. **Multistore layout is part of state commitment**: each KVStore is included in `CommitInfo.StoreInfos`, and the final `appHash` is a Merkle root over all store commit hashes. Even an empty IAVL store has a non-nil hash, so adding one changes the app hash immediately.
2. **Store mount timing constraint**: `MountStoreWithDB` can only be called before `LoadLatestVersion()` (at app startup). The multistore topology is immutable once loaded. The upgrade handler runs during block execution and cannot mount new stores.
3. **Fork-based upgrade requires continuous execution**: the chain runs from genesis through the upgrade height without halting. The upgrade handler needs the DKG store to be mounted at height H, so it must be mounted at app startup — which changes the app hash from the first commit.

Therefore, all validators must perform a **coordinated binary upgrade** (stop and restart with the new binary) before the upgrade height. The new binary will have a different app hash from the first block it processes, making mixed old/new binary operation impossible. Cosmovisor would be helpful for this upgrade. 

## Suggested solution

[...]

PR #693, opened 2026-03-07
feat(dkg): implement upgrade handler for dkg activation

0xHansLee:
Implement the v2.0.0 upgrade handler to activate the DKG module on live chains and ensure all DKG functionality is properly gated behind the upgrade height.

  ### Changes

  - **v2.0.0 upgrade handler** (`client/app/upgrades/v_2_0_0/`): `RunMigrations` for DKG InitGenesis, set default DKG params, enable vote extensions at `upgradeHeight + 1`
  - **DKG height gating** (`client/x/dkg/keeper/`): `BeginBlocker`, `ExtendVote`, `VerifyVoteExtension`, and `PrepareVotes` return no-ops before v2.0.0 activation
  - **Proposal compatibility** (`client/app/prouter.go`, `client/x/evmengine/keeper/abci.go`): `PrepareVotes` returns `nil` (not empty `MsgAddDkgVote`) before `v200Height + 2`; `ProcessProposal` conditionally expects `MsgAddDkgVote`; `PrepareProposal` only includes it when non-nil
  - **Store loader** (`client/app/upgrades.go`): `UpgradeStoreLoader` pre-adds stores from all upgrades (past and future) so the binary starts cleanly at any height
  - **UBI distribution** (`client/x/evmstaking/keeper/ubi.go`): skip DKG committee reward distribution before v2.0.0
  - **Network config** (`lib/netconf/upgrades.go`): add V200 constant and upgrade heights for localnet/aeneid

  ### Why coordinated binary upgrade, not rolling upgrade?

  We investigated whether validators could perform a rolling upgrade — swapping binaries one at a time before the upgrade height while maintaining consensus with old-binary validators. **This is not possible** due to three Cosmos SDK constraints:

  1. **Multistore layout = state commitment.** The `appHash` is a Merkle root over all KVStore commit hashes in `CommitInfo`. Adding the `dkg` KVStore — even empty — changes this Merkle root. The new binary produces a different `appHash` from its very first commit, breaking consensus with old-binary validators.

  2. **Store mount is startup-only.** `MountStoreWithDB` must be called before `LoadLatestVersion()`. After the multistore loads, its topology is frozen. The upgrade handler runs mid-block and cannot mount new stores. So the DKG store must be mounted at app startup, before the upgrade height.

  3. **Fork-based upgrade = continuous execution.** The chain doesn't halt at the upgrade height. The upgrade handler needs the DKG store already mounted to run `InitGenesis`. Since the store must be mounted at startup and mounting changes the app hash, there's no window where old and new binaries can coexist.

  **Deployment strategy**: all validators must stop and restart with the new binary before the upgrade height. This is a coordinated upgrade, not a rolling one.

  ### Vote extension timing

  Vote extensions require 2 blocks to propagate after enablement:
  - Upgrade at height H: handler sets `vote_extensions_enable_height = H + 1`
  - Height H+1: CometBFT starts collecting vote extensions
  - Height H+2: vote extensions appear in `LocalLastCommit`, `MsgAddDkgVote` included in proposals

  ## Test plan

  - [x] `client/x/dkg/keeper` — unit tests pass (31.7% coverage)
  - [x] `client/app` — `TestProcessProposalRouter` passes with v2.0.0 height gating
  - [x] `client/x/evmstaking/keeper` — UBI tests pass
  - [x] 4-validator localnet test:
    - All validators start with new binary
    - Upgrade activates at height 110
    - DKG module starts, vote extensions collected after height 112
    - Blocks continue producing normally post-upgrade

issue: #692

### DIFF
--- a/client/app/prouter.go
+++ b/client/app/prouter.go
@@ -18,6 +18,7 @@ import (
 	evmenginetypes "github.com/piplabs/story/client/x/evmengine/types"
 	"github.com/piplabs/story/lib/errors"
 	"github.com/piplabs/story/lib/log"
+	"github.com/piplabs/story/lib/netconf"
 )
 
 // processTimeout is the maximum time to process a proposal.
@@ -75,7 +76,16 @@ func makeProcessProposalHandler(router *baseapp.MsgServiceRouter, txConfig clien
 		// Ensure only expected messages types are included the expected number of times.
 		expectedMsgCounts := map[string]int{
 			sdk.MsgTypeURL(&evmenginetypes.MsgExecutionPayload{}): 1, // Only a single EVM execution payload is allowed.
-			sdk.MsgTypeURL(&dkgtypes.MsgAddDkgVote{}):             1,
+		}
+
+		// MsgAddDkgVote is only expected after v2.0.0 activation (+2 blocks
+		// for vote extensions to propagate through LocalLastCommit).
+		v200Height, v200Err := netconf.GetUpgradeHeight(ctx.ChainID(), netconf.V200)
+		if v200Err != nil {
+			return rejectProposal(ctx, errors.Wrap(v200Err, "get v2.0.0 upgrade height"))
+		}
+		if req.Height > v200Height+1 {
+			expectedMsgCounts[sdk.MsgTypeURL(&dkgtypes.MsgAddDkgVote{})] = 1
 		}
 
 		rawTX := req.Txs[0]
--- a/client/app/upgrades.go
+++ b/client/app/upgrades.go
@@ -2,7 +2,9 @@ package app
 
 import (
 	"fmt"
+	"sort"
 
+	"cosmossdk.io/store/rootmulti"
 	storetypes "cosmossdk.io/store/types"
 	upgradetypes "cosmossdk.io/x/upgrade/types"
 
@@ -15,6 +17,7 @@ import (
 	"github.com/piplabs/story/client/app/upgrades/singularity/virgil"
 	"github.com/piplabs/story/client/app/upgrades/terence"
 	"github.com/piplabs/story/client/app/upgrades/v_1_2_0"
+	"github.com/piplabs/story/client/app/upgrades/v_2_0_0"
 	"github.com/piplabs/story/lib/errors"
 	"github.com/piplabs/story/lib/netconf"
 )
@@ -28,6 +31,7 @@ var (
 		polybius.Upgrade,
 		terence.Upgrade,
 		horace.Upgrade,
+		v_2_0_0.Upgrade,
 	}
 	// Forks are for hard forks that breaks backward compatibility.
 	Forks = []upgrades.Fork{
@@ -36,6 +40,7 @@ var (
 		polybius.Fork,
 		terence.Fork,
 		horace.Fork,
+		v_2_0_0.Fork,
 	}
 )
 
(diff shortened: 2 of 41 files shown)


lib/netconf/upgrades.go
```diff
@@ -52,6 +56,7 @@ var UpgradeHistories = map[string]UpgradeMap{
 		Terence:  10886688,
 		V142:     12088950,
 		Horace:   14017000,
+		V200:     100000000, // TBD: set before deployment
```
ramtinms, line 59: do we need another PR for this ?

ramtinms (approved):

lib/netconf/upgrades.go
0xHansLee: Yes, need to set the target height for Aeneid. Also need the name for this.

client/app/upgrades.go
```diff
@@ -75,12 +80,76 @@ func (a *App) setupUpgradeStoreLoaders() {
...
-			if len(storeUpgrades.Renamed) > 0 || len(storeUpgrades.Deleted) > 0 || len(storeUpgrades.Added) > 0 {
-				return ms.LoadLatestVersionAndUpgrade(&storeUpgrades)
+		lastVersion := ms.LastCommitID().Version
+		nextVersion := lastVersion + 1
+
+		// On a fresh genesis chain (no commits yet), all modules are already
+		// registered and their stores are mounted. Applying store upgrades
+		// would conflict with existing stores (e.g., trying to add DKG store
+		// that already exists from genesis). Skip the upgrade loader entirely.
+		if lastVersion == 0 {
```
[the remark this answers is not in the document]
0xHansLee: This is not a valid concern. When lastVersion == 0, it's a fresh genesis chain where no blocks have been committed yet. In this case:

1. ALL modules (including DKG) are already registered in `app_config.go`, so their stores are already mounted by the module manager before LoadLatestVersion is called.
2. No upgrades have ever executed, so there are no store operations to apply.
3. `storeUpgradesMap` entries all have height > 0, which are future upgrades — but on a fresh genesis chain, those modules' stores already exist from genesis.

`StoreUpgrades` are only needed when upgrading an existing chain (e.g., v1.5.3 → v2.0.0) where the old binary didn't have the DKG store. In that case lastVersion > 0, so the early return is not hit.

Applying `StoreUpgrades` on a fresh genesis chain would actually cause errors — trying to add a store that's already mounted results in "initial version set to X, but found earlier version Y" conflicts.

client/app/prouter.go
```diff
@@ -75,7 +76,16 @@ func makeProcessProposalHandler(router *baseapp.MsgServiceRouter, txConfig clien
...
+		}
+
+		// MsgAddDkgVote is only expected after v2.0.0 activation (+2 blocks
+		// for vote extensions to propagate through LocalLastCommit).
+		v200Height, v200Err := netconf.GetUpgradeHeight(ctx.ChainID(), netconf.V200)
+		if v200Err != nil {
+			return rejectProposal(ctx, errors.Wrap(v200Err, "get v2.0.0 upgrade height"))
+		}
+		if req.Height > v200Height+1 {
+			expectedMsgCounts[sdk.MsgTypeURL(&dkgtypes.MsgAddDkgVote{})] = 1
```
jdubpark, line 88: Do we expect `MsgAddDkgVote` for every block after the upgrade height (+2)? This will halt the node if arrived msg count of MsgAddDkgVote is 0 (or >1)

0xHansLee: 0 of MsgAddDkgVote would not be a matter, but if the number of msg is greater than 1, it will be rejected ([ref](https://github.com/piplabs/story/blob/41e5271c13a75d04e03a1be6d5aefd3464defed4/client/app/prouter.go#L88)). In theory, the msg can be included from the upgrade height + 1.

PR #696, opened 2026-03-11
feat(dkg): distribute cdr collected fee to committee members

stevemilk:
Summary
- Route CDR fees into a dedicated fee pool and distribute to validators on DKG finalize
- Track partial decryption submit counts per validator (reset each round)
- Refund submit fees from the pool; emit CDR fee collected events for observability

issue: none


contracts/src/protocol/CDR.sol
```diff
@@ -11,6 +11,11 @@ import { ICDRWriteCondition } from "../interfaces/ICDRWriteCondition.sol";
 import { ICDRReadCondition } from "../interfaces/ICDRReadCondition.sol";
 
 contract CDR is ICDR, Ownable2StepUpgradeable, ReentrancyGuardUpgradeable, PausableUpgradeable, UUPSUpgradeable {
+    uint8 private constant FEE_TYPE_ALLOCATE = 0;
+    uint8 private constant FEE_TYPE_WRITE = 1;
+    uint8 private constant FEE_TYPE_READ = 2;
+    uint8 private constant FEE_TYPE_SUBMIT_PARTIAL = 3;
```
0xHansLee: How about using enum?

0xHansLee (approved):

contracts/src/protocol/CDR.sol
stevemilk: yeah we can

PR #700, opened 2026-03-12
feat(dkg): add a new rpc for users to get partials

stevemilk:
Changes:
- users are able to fetch partials which are grouped by round according to vaultUUID+pubKey
- Todo: need to discuss pruning of stored partials, like how long we should keep the partials for

issue: none


client/proto/story/dkg/v1/types/query.proto
```diff
@@ -42,6 +42,10 @@ service Query {
   rpc GetLatestActiveDKGNetwork(QueryGetLatestActiveDKGNetworkRequest) returns (QueryGetLatestActiveDKGNetworkResponse) {
     option (google.api.http).get = "/story/dkg/v1/latest_active";
   }
+
+  rpc GetCDRPartials(QueryGetCDRPartialsRequest) returns (QueryGetCDRPartialsResponse) {
+    option (google.api.http).get = "/story/dkg/v1/partial_decrypt/{requester_pub_key_hex}/{uuid}";
```
0xHansLee: How about we change the order, `{uuid}/{requester_pub_key_hex}`? Because the `uuid` is higher-level concept.

client/proto/story/dkg/v1/types/query.proto
```diff
@@ -155,4 +159,21 @@ message QueryGetLatestActiveDKGNetworkResponse {
...
-}
\ No newline at end of file
+}
+
+message QueryGetCDRPartialsRequest {
+  string requester_pub_key_hex = 1 [
+    (gogoproto.moretags) = "yaml:\"requester_pub_key_hex\""
+  ];
+
+  uint32 uuid = 2 [
```
0xHansLee: How about `label` here for the consistency in CDR? Let's use `uuid` in the contract only.

client/x/dkg/keeper/keeper.go
```diff
@@ -58,8 +58,9 @@ type Keeper struct {
 	SettlementBalance  collections.Item[string]                         // remaining UBI after committee distribution during FinalizeDKGRound
 	KernelUpgradeInfos collections.Map[string, types.KernelUpgradeInfo] // key: upgradeVersion
 
-	DKGPartialDecrypt      collections.Map[string, []byte] // key: round_validator_pid_labelHash
-	DecryptRequestRegistry collections.Map[string, uint64] // key: requesterPubKeyHash_labelHash; value: blockHeight when request was registered
+	// TODO: need to prune historical partial decryptions and decrypt request registry entries after some retention period to prevent unbounded growth. Can be done in EndBlocker.
```
0xHansLee: I don't think we need to prune the partials. It's enough to use pruning provided by cometbft.

client/x/dkg/keeper/query.go
```diff
@@ -126,3 +131,64 @@ func (k *Keeper) GetLatestActiveDKGNetwork(ctx context.Context, request *types.Q
...
+
+	prefix := dkgPartialDecryptPrefix(requesterPubKey, label[:])
+	rangePrefix := (&collections.Range[string]{}).Prefix(prefix)
+	iter, err := k.DKGPartialDecrypt.Iterate(ctx, rangePrefix)
+	if err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+	defer iter.Close()
+
+	grouped := make(map[uint32][]types.DKGPartialDecryptionSubmission)
```
0xHansLee, line 156: Why the `round` is used for key here? The partials are based on the round when the `read` request is submitted, so I assume that every partials will have the same round. Am I missing something?

client/proto/story/dkg/v1/types/query.proto
stevemilk: Users call `read` function in cdr contract with `uuid`. They don't actually know `label`, so i think better use `uuidd` here

client/x/dkg/keeper/query.go
stevemilk: Users can read the same vault in different rounds so maybe each round there'll be some partials collected.

client/x/dkg/keeper/keeper.go
stevemilk: I think in most case we can rely on cometbft pruning as long as the speed of state growth has a up limit.
 But for this case, the number of DecryptRequestRegistry is not controlled by the protocol but by users. This is an attack vector, though not that severe.

0xHansLee: I don't think it's not a matter. User also can make penny stake state, which consumes the storage of all nodes as well (with lowering the min stake amount). 

And I think the encrypted partials should be permanent on chain, so that the eligible reader can fetch the data anytime.

client/x/dkg/keeper/query.go
0xHansLee: The partials would be based on the round when the request is submitted. That means, one request, whose key is composed of requesterPubKey and label, will have partials from a single round. All partials should be generated and submitted from the members of the same round.

0xHansLee (approved):

client/x/dkg/keeper/query.go
stevemilk: Vault data can be updated, we should allow one user `read` multiple times in different rounds, maybe even in the same round, need to make it clearer later.

PR #705, opened 2026-03-13
fix(dkg): fix issues found in e2e testing

0xHansLee:
## Summary

  Bug fixes and improvements discovered during end-to-end DKG testing on devnet.
  Each commit addresses a specific issue found during multi-validator DKG round execution.

  ### Issues resolved

  - **Async goroutine accessing Cosmos KV store** (`pre-compute old code commitment`):
    `getOldCodeCommitment()` was called inside an async goroutine using `context.Background()`,
    which cannot access the Cosmos KV store. Moved the call before the goroutine while SDK
    context is still available. Also fixed nil pointer when no previous active round exists.

  - **Async goroutine building dealer pub key map** (`move buildDealerPubKeyMap`):
    Same class of bug — `buildDealerPubKeyMap()` reads DKG registrations from the KV store
    but was called inside the async justification-processing goroutine. Pre-computed before
    goroutine launch.

  - **Finalization rejected for valid dealers** (`fix finalization check order`):
    `validateParticipantsRoot` was called before the double-finalization guard and dealer
    invalidation check. A dealer who was invalidated mid-round would hit the participants
    root validation (which includes all registered participants) before being rejected by
    the invalidation check, producing a confusing error. Moved `validateParticipantsRoot`
    after the guard checks. Also removed unused `isInPrevActiveValSet` and `advanceBlock`
    helper.

  - **Stale enclave type after upgrade resharing** (`resolve enclave type for all DKG rounds`):
    After an upgrade resharing completes, `k.enclaveType` still holds the OLD enclave type.
    Subsequent non-upgrade rounds would register with a stale enclave type, causing
    finalization signature mismatches. Added `ResolveEnclaveType()` that queries on-chain
    whitelisted enclave types by code commitment, and resolves the correct type for every
    round (not just upgrades).

  - **MsgAddDkgVote authority mismatch** (`address dkg code review findings`):
    `PrepareVotes` hardcoded `authtypes.NewModuleAddress(types.ModuleName)` as the message
    authority instead of using `k.GetAuthority()`, which is set during keeper initialization.
    This caused authority mismatch when the keeper's authority differed from the default
    module address.

  - **Upgrade resharing and DKG lifecycle improvements** (`implement upgrade resharing`):
    Implements kernel upgrade support with dual-binary routing, upgrade-aware dealing that
    routes old members to the old kernel and new members to the new kernel, and adds
    comprehensive lifecycle integration tests (758 lines) covering full DKG rounds across
    registration → dealing → finalization → active transitions.

issue: none

### DIFF
--- a/client/x/dkg/keeper/contract_client.go
+++ b/client/x/dkg/keeper/contract_client.go
@@ -151,7 +151,15 @@ func (c *ContractClient) Register(ctx context.Context, round uint32, enclaveType
 		return nil, errors.Wrap(err, "failed to pack register call data")
 	}
 
-	return c.sendWithRetry(ctx, "Register", c.dkgContractAddr, callData, func(auth *bind.TransactOpts) (*types.Transaction, error) {
+	// Query the registration fee from the DKG contract
+	fee, err := c.dkgContract.Fee(nil)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to query DKG fee")
+	}
+
+	log.Info(ctx, "DKG registration fee queried", "fee_wei", fee.String())
+
+	return c.sendWithRetry(ctx, "Register", c.dkgContractAddr, callData, fee, func(auth *bind.TransactOpts) (*types.Transaction, error) {
 		return c.dkgContract.Register(auth, enclaveReport, enclaveInstanceData, startBlockHeightBig, startBlockHash32, []byte{})
 	})
 }
@@ -185,7 +193,15 @@ func (c *ContractClient) Finalize(
 		return nil, errors.Wrap(err, "failed to pack finalize call data")
 	}
 
-	return c.sendWithRetry(ctx, "Finalize", c.dkgContractAddr, callData, func(auth *bind.TransactOpts) (*types.Transaction, error) {
+	// Query the DKG fee — the contract requires it for finalization as well.
+	fee, err := c.dkgContract.Fee(nil)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to query DKG fee for finalize")
+	}
+
+	log.Info(ctx, "DKG finalization fee queried", "fee_wei", fee.String())
+
+	return c.sendWithRetry(ctx, "Finalize", c.dkgContractAddr, callData, fee, func(auth *bind.TransactOpts) (*types.Transaction, error) {
 		return c.dkgContract.Finalize(auth, round, c.fromAddress, enclaveType, participantsRoot32, globalPubKey, publicCoeffs, pubKeyShare, signature)
 	})
 }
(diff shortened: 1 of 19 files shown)


client/x/dkg/keeper/dkg_svc.go
```diff
@@ -11,6 +11,48 @@ import (
...
+					log.Info(context.Background(), "Superseded stale DKG lock from previous round",
+						"old_round", current,
+						"new_round", r,
+					)
+				}
+
+				return true
+			}
+
+			continue // CAS failed, retry
```
jdubpark, line 45: Could there be an edge case where this continues to fail and get stuck into forever loop? In that case, we might need max retry for the loop.

client/x/dkg/keeper/dkg_svc_dealing.go
```diff
@@ -136,8 +137,14 @@ func (k *Keeper) handleDKGProcessDeals(ctx context.Context, dkgNetwork *types.DK
...
 	}
 
-	if session.Phase != types.PhaseDealing {
-		log.Warn(ctx, "Session not in dealing phase, skipping process deals", nil,
+	// Accept deals in both Initialized and Dealing phases.
+	// Deals from other validators arrive via vote extensions as soon as the DKG stage
+	// transitions to Dealing. However, the local session only transitions from
+	// Initialized to Dealing after GenerateDeals completes (which can take ~30s).
+	// If we reject deals during Initialized phase, they are permanently lost because
+	// vote extensions deliver each deal exactly once.
```
jdubpark, line 145: good catch 🙏

client/x/dkg/keeper/dkg_svc_dealing.go
```diff
@@ -160,7 +167,9 @@ func (k *Keeper) handleDKGProcessDeals(ctx context.Context, dkgNetwork *types.DK
 		}
 
 		for _, deal := range deals {
-			if deal.RecipientIndex == session.Index {
+			// RecipientIndex is 0-based (Kyber), session.Index is 1-based (on-chain).
+			// Guard against unset index (0) to avoid uint32 underflow.
+			if session.Index > 0 && deal.RecipientIndex == session.Index-1 {
```
jdubpark, line 172: seems like the above bot comments are missing this conditional check (`session.Index > 0`) 🍭

jdubpark (approved):

client/x/dkg/keeper/dkg_handler.go
```diff
@@ -161,18 +161,28 @@ func (k *Keeper) Finalized(ctx context.Context, round uint32, msgSender common.A
...
+
+	allRegs := append(verifiedRegs, finalizedRegs...)
+
+	if len(allRegs) == 0 {
+		return errors.New("no verified or finalized DKG registrations found")
 	}
 
-	addrs := make([]string, 0, len(verifiedRegs))
-	for _, reg := range verifiedRegs {
+	addrs := make([]string, 0, len(allRegs))
```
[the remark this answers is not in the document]
0xHansLee: A validator cannot appear in both Verified and Finalized lists simultaneously. The status transitions are: Verified → Finalized (one-way). When a validator finalizes, its status changes FROM Verified TO Finalized. The query fetches registrations by status, so each validator appears in exactly one list. No deduplication is needed.

client/x/dkg/keeper/dkg_svc.go
0xHansLee: Good point. While in practice CAS contention resolves in 1-2 iterations (only 1-2 goroutines contend per block, and a successful CAS removes the contender), the unbounded loop is not ideal from a defensive programming perspective. Added a max retry bound (10 attempts) — if exhausted, returns false (treat as failed acquisition). This is purely defensive; the bound should never be reached in normal operation.

client/x/dkg/keeper/contract_client.go
```diff
@@ -218,13 +234,14 @@ func (c *ContractClient) SubmitEncryptedPartialDecryption(
 		return nil, errors.Wrap(err, "failed to pack submitEncryptedPartialDecryption call data")
 	}
 
-	return c.sendWithRetry(ctx, "SubmitEncryptedPartialDecryption", c.cdrContractAddr, callData, func(auth *bind.TransactOpts) (*types.Transaction, error) {
+	return c.sendWithRetry(ctx, "SubmitEncryptedPartialDecryption", c.cdrContractAddr, callData, nil, func(auth *bind.TransactOpts) (*types.Transaction, error) {
```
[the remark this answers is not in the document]
0xHansLee: Good catch — this is a real bug. The CDR contract's `submitEncryptedPartialDecryption` requires baseFee payment via `_collectFee()`, but the story client was passing nil (0 ETH). Fixed in [41e5271](https://github.com/piplabs/story/pull/705/commits/41e5271c13a75d04e03a1be6d5aefd3464defed4) by querying `cdrContract.BaseFee()` and passing it as the transaction value, matching the pattern used by `Register` and `Finalize`.

stevemilk (approved):

Issue #703, opened 2026-03-13
Duplicate DKG registration on resume/retry corrupts participant index


barry-peng-story:
# Description

On resume/retry (when `Phase=Failed` or after `UpdateSession` failure),
`handleDKGRegistration` runs again and calls the contract `Register()` without
checking whether the validator is already registered for the current round.
This causes duplicate `Registered` events that overwrite the stored registration
and corrupt the participant index (`DKGRegistration.Index`).

### 1. `ResumeDKGService` unconditionally restarts registration

`story/client/x/dkg/keeper/dkg_svc.go:44-60`

```go
case types.DKGStageRegistration:
    session.UpdatePhase(types.PhaseInitializing)
    if err := k.stateManager.UpdateSession(ctx, session); err != nil { ... }
    go func() {
        k.handleDKGRegistration(asyncCtx, dkgNetwork)  // re-runs full registration
    }()
```

When `Phase=Failed` and `Stage=Registration`, `ResumeDKGService` resets the
phase to `PhaseInitializing` and re-invokes `handleDKGRegistration` with no
check for prior on-chain registration.

### 2. `handleDKGRegistration` always creates a new session

`story/client/x/dkg/keeper/dkg_svc_registration.go:42-64`

```go
// Always constructs a brand-new session — never loads the existing one
session := types.NewDKGSession(dkgNetwork.Round, dkgNetwork.ActiveValSet, ...)
if err := k.stateManager.CreateSession(ctx, session); err != nil { ... }
```

There is no check for whether a session already exists with
`Phase >= PhaseInitialized`.

### 3. `callContractRegister` has no idempotency guard

`story/client/x/dkg/keeper/dkg_svc_registration.go:225-249`

```go
func (k *Keeper) callContractRegister(ctx context.Context, session *types.DKGSession) error {
    // No pre-check — calls Register() unconditionally
    if _, err := k.contractClient.Register(...); err != nil {
        return err
    }
    return nil
}
```

Compare with `callTEEGenerateAndSealKey` (lines 116–120), which already has an
idempotency guard:

```go
if len(session.DKGPubKey) > 0 && len(session.CommPubKey) > 0 && len(session.EnclaveReport) > 0 {
    log.Info(ctx, "Already generated and sealed the key, skipping GenerateAndSealKey request")
    return nil
}
```

`callContractRegister` has no equivalent check.

### 4. On-chain `getNextDKGRegistrationIndex` does not deduplicate

[...]

PR #715, opened 2026-03-18
fix(dkg): add defense in depth guard against duplicate registration

0xHansLee:
## Summary

  - Add defense-in-depth guard against duplicate DKG registration in `handleDKGRegistration` by passing a pre-computed `alreadyRegistered` flag from the SDK context into the async goroutine
  - When the validator is already registered on-chain, `callContractRegister` is skipped and the session advances directly to `PhaseInitialized`, preventing key material from being overwritten with different keys
  - Both callers (`InitiateDKGRound` and `resumeFailedSession`) now pre-compute the registration check in the SDK context and pass it to the goroutine, since the async context cannot access the Cosmos KV store

issue: #703

### DIFF
--- a/client/x/dkg/keeper/dkg_initialization.go
+++ b/client/x/dkg/keeper/dkg_initialization.go
@@ -80,16 +80,15 @@ func (k *Keeper) InitiateDKGRound(ctx context.Context, isUpgrade bool) error {
 	}
 
 	if k.isDKGSvcEnabled {
-		// Skip registration if this validator already has an on-chain registration
-		// for this round. This prevents overwriting a valid registration with
-		// different keys after a reset where sealed_keys were deleted.
-		if k.isAlreadyRegistered(ctx, roundNum) {
+		// Pre-compute registration check while we still have SDK context.
+		// The async goroutine uses context.Background() which cannot access
+		// the Cosmos KV store.
+		alreadyRegistered := k.isAlreadyRegistered(ctx, roundNum)
+		if alreadyRegistered {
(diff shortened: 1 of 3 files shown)


jdubpark (approved):

client/x/dkg/keeper/dkg_initialization.go
```diff
@@ -80,24 +80,23 @@ func (k *Keeper) InitiateDKGRound(ctx context.Context, isUpgrade bool) error {
...
 
 	if k.isDKGSvcEnabled {
-		// Skip registration if this validator already has an on-chain registration
-		// for this round. This prevents overwriting a valid registration with
-		// different keys after a reset where sealed_keys were deleted.
-		if k.isAlreadyRegistered(ctx, roundNum) {
+		// Pre-compute registration check while we still have SDK context.
+		// The async goroutine uses context.Background() which cannot access
+		// the Cosmos KV store.
+		alreadyRegistered := k.isAlreadyRegistered(ctx, roundNum)
```
barry-peng-story, line 86: didn't see the isAlreadyRegistered implementation, is it in this branch?

0xHansLee: It's implemented in the prev PR. It's in this file, https://github.com/piplabs/story/pull/715/changes#diff-567dff34984dfddfe6f44207dd56efb843ec38d9fb5958c9f1835c829319074dR109

stevemilk (approved):

PR #727, opened 2026-03-22
fix: dkg decrypt worker lifecycle and stale request handling

jdubpark:
## Summary

- **Fix decrypt worker premature exit**: The worker was using the caller's short-lived async context, which gets cancelled when the parent goroutine exits. Switched to `context.Background()` so the worker runs for the lifetime of the process.
- **Ensure decrypt worker starts on resume**: Added a check in `ResumeDKGService` to start the decrypt worker when the session is completed and the DKG round is active. Covers node restarts and cases where the worker was never started due to context cancellation.
- **Also start worker on re-entry into `handleDKGComplete`**: Ensures the worker is running even if completion was already processed (e.g., after node restart).
- **Drop decrypt requests for unavailable kernels**: When old events are replayed during chain catch-up after a kernel binary change, the sealed keys are unreachable. These requests are now detected via `kernelRouter.GetClient` and dropped with a warning log instead of failing repeatedly.
- **Add operational logging**: Log when the worker starts, when it processes a queue, and when individual decrypt requests succeed.
- **Regenerated contract bindings** to stay in sync.

issue: none


client/x/dkg/keeper/dkg_svc.go
```diff
@@ -237,25 +247,28 @@ func (k *Keeper) resumeFailedSession(ctx context.Context, session *types.DKGSess
 
 // StartDecryptWorker launches a background loop (non-ABCI) that drains pending decrypt requests
 // and performs TDH2 partial decrypts. Only one worker runs.
-func (k *Keeper) StartDecryptWorker(ctx context.Context) {
+// The worker uses its own long-lived context (derived from context.Background) because the
+// caller's context (dkgAsyncContext) is short-lived and gets cancelled when the parent
+// goroutine exits. The decrypt worker must run for the lifetime of the process.
+func (k *Keeper) StartDecryptWorker(_ context.Context) {
```
0xHansLee: We may be able to remove the arg here

client/x/dkg/keeper/dkg_svc.go
```diff
@@ -269,6 +282,31 @@ func (k *Keeper) processDecryptQueue(ctx context.Context) {
 			continue
 		}
 
+		// Skip sessions whose kernel binary is no longer connected.
+		// This happens when old events are replayed during chain catch-up
+		// after a kernel binary change — the sealed keys are unreachable.
+		if _, err := k.kernelRouter.GetClient(session.CodeCommitment); err != nil {
```
0xHansLee: How about using k.getClientWithReconnect() here?

stevemilk:
It occurs to me we should also drop stale requests during resync even there's no binary change for kernel.
Decryption requests have a 200-block timeout, so validators don’t need to process requests from blocks with a height lower than `chainTip-200`. How do you think?

client/x/dkg/keeper/dkg_svc.go
jdubpark: Yes, changed in 57df4cc

client/x/dkg/keeper/dkg_svc.go
jdubpark: Yes, changed in 57df4cc

jdubpark:
> It occurs to me we should also drop stale requests during resync, even if there's no binary change for kernel. Decryption requests have a 200-block timeout, so validators don’t need to process requests from blocks with a height lower than `chainTip-200`. How do you think?

Yes, good idea. I faced a similar issue when resetting. Old decryption requests got cached in the store with older round and code commitment data. For example, even when I reset story-kernel to round 1 and a new code commitment, an old decryption request would try to process as round 6 and an outdated code commitment. Had to skip them.

Added as a DKG module param in a3bbc58. @0xHansLee @stevemilk

stevemilk (approved):

0xHansLee (approved):
Please fix the failed gha and merge it.

PR #731, opened 2026-03-24
feat: license holder read condition contract

jdubpark:
### Summary
Add `LicenseReadCondition` contract that gates CDR vault (registered as IP asset) read access based on ownership of a valid, non-revoked Story Protocol license token for the vault's IP asset. Include Foundry tests with a mock LicenseToken contract.

### Details
The LicenseReadCondition implements ICDRReadCondition and verifies that the caller owns at least one license token whose licensorIpId matches the vault's IP asset. Revoked tokens are rejected.

issue:

### DIFF
--- a/contracts/src/protocol/cdr/LicenseReadCondition.sol
+++ b/contracts/src/protocol/cdr/LicenseReadCondition.sol
@@ -1,0 +1,64 @@
+// SPDX-License-Identifier: GPL-3.0-only
+pragma solidity 0.8.23;
+
+import { ICDRReadCondition } from "../../interfaces/ICDRReadCondition.sol";
+
+/// @notice Minimal interface for Story Protocol's LicenseToken (ERC721-based).
+interface ILicenseToken {
+    struct LicenseTokenMetadata {
+        address licensorIpId;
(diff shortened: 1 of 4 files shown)


contracts/src/protocol/cdr/LicenseReadCondition.sol
```diff
@@ -0,0 +1,64 @@
...
+    /// @param caller The address attempting to read the vault
+    /// @return True if the caller owns at least one valid license token for the IP asset
+    function checkReadCondition(
+        uint32,
+        bytes calldata accessAuxData,
+        bytes calldata readConditionData,
+        address caller
+    ) external view override returns (bool) {
+        (address licenseToken, address ipId) = abi.decode(readConditionData, (address, address));
+        uint256[] memory tokenIds = abi.decode(accessAuxData, (uint256[]));
```
0xHansLee, line 44: qq. Should we allow multiple tokens here?

jdubpark: Yes I think we should, let me know if there are downsides to this

PR #755, opened 2026-03-25
fix(dkg): resolve issues reported by the AI audits

### LINKED ISSUE #745: CDR-078 [P3]: KernelRouter.Disconnect doesn't remove ccByEP entry
## Audit Finding CDR-078

| Field | Value |
|-------|-------|
| **Severity** | P3 — Stale State |
| **Confidence** | 95 |
| **Location** | `client/x/dkg/keeper/kernel_router.go:162-174` |

### Summary

`Disconnect` removes the client from `r.clients` and `r.closers` but does NOT remove the corresponding `r.ccByEP[endpoint]` entry. This means `disconnectedEndpoints()` (which checks `r.ccByEP`) will never return the endpoint again, preventing `TryReconnect` from reconnecting after a disconnect.

### Details

```go
// kernel_router.go:162-174 — missing ccByEP cleanup
func (r *KernelRouter) Disconnect(codeCommitment []byte) {
    r.mu.Lock()
    defer r.mu.Unlock()

    codeCommitmentHex := hex.EncodeToString(codeCommitment)
    if closer, ok := r.closers[codeCommitmentHex]; ok {
        _ = closer.Close()
        delete(r.closers, codeCommitmentHex)
    }
    delete(r.clients, codeCommitmentHex)
    // BUG: r.ccByEP still maps endpoint -> codeCommitmentHex
}
```

`disconnectedEndpoints()` returns endpoints NOT in `ccByEP`:
```go
func (r *KernelRouter) disconnectedEndpoints() []string {
    for _, ep := range r.endpoints {
        if _, ok := r.ccByEP[ep]; !ok {
            out = append(out, ep)
        }
    }
}
```

Since `ccByEP[ep]` still exists after `Disconnect`, the endpoint never appears in the disconnected list.

### Impact

After a kernel disconnection, the CL cannot automatically reconnect to that endpoint. Manual node restart required.

### Recommendation

```go
// Add to Disconnect, inside the lock:
for ep, cc := range r.ccByEP {
    if cc == codeCommitmentHex {
        delete(r.ccByEP, ep)
    }
}
```

### Characterization Test

Branch: [`audit/CDR-078`](https://github.com/piplabs/story/tree/audit/CDR-078)
`client/x/dkg/keeper/kernel_router_internal_test.go:256`
Test: `TestTODO_CDR078_Disconnect_DoesNotRemoveCcByEP`

The test connects an endpoint, disconnects it, then checks `disconnectedEndpoints()` — it returns empty (should return the endpoint), proving the stale `ccByEP` entry.

0xHansLee:
## Summary

  Remediation of findings from the CDR AI Security Audit Report and Apex Report.

  ## Audit Finding Resolution

  ### Contracts (`fix(contracts)` — `79044711`)

  | Finding | Resolution |
  |---------|-----------|
  | SC-03 | Add input validation for enclave report length (`> 0`) in SGXValidationHook |
  | SC-06 | Add max encrypted partial size (1024 bytes) in CDR.submitEncryptedPartialDecryption |
  | SC-07 | Mark extraction functions as `internal pure` in CDR and SGXValidationHook |
  | SC-08 | Add `nonReentrant` to CDR.submitEncryptedPartialDecryption, CDR.allocate, DKG.register, DKG.finalize |
  | SC-09 | Add `nonReentrant` to CDR.allocate |
  | SC-12 | Use `storage` pointer instead of `memory` for Vault in CDR.write and CDR.read |
  | L1-02 | Add max encrypted data size (1 KB) validation in CDR.allocate |
  | L1-04 | Add documentation for DKG.scheduleUpgrade version format (opaque identifier, not semver) |
  | STOR-11 | Add `require(validatorAddr == msg.sender)` in DKG.finalize to prevent single TEE keypair finalizing for multiple validators |
  | CDR-006 | Add documentation for intentional condition bypass when msg.sender is the condition contract |
  | CDR-018 | Add UUID uint32 overflow check in CDR.allocate |
  | SC-04 | Add documentation: zero fees intentionally allowed for testing/early deployment |
  | SC-10 | Add documentation: read access control enforced through readConditionAddr contract |

  ### DKG Keeper (`fix(dkg)` — `8dc670f1`)

  | Finding | Resolution |
  |---------|-----------|
  | STOR-7 / STOR-10 | Remove all CL-side deduplication (deals, responses, justifications). CL cannot validate DKG message authenticity — only the kernel can. First-seen dedup on unsigned fields allows suppression attacks. Kernel handles dedup. |
  | STOR-14 | Remove recipient_index filtering; pass all justifications to kernel |
  | STOR-24 | Remove MaxJustificationsPerBlock truncation. VerifyVoteExtension already caps items per VE. |
  | STOR-2 | Add fallback for new validators in upgrade resharing rounds (no old code commitment → use first available kernel client) |
  | CDR-004 | Add documentation clarifying that partial decryption signatures consistently omit the Ethereum Signed Message prefix |
  | CDR-017 | Use `NewIntFromUint64` to prevent overflow in CDR fee distribution |
  | CDR-020 | Align justification error handling with deal/response pattern (return on kernel error instead of continue) |
  | L1-06 | Add documentation for intentionally permissive ABCI++ proposal pattern |
  | — | Remove Ethereum Signed Message prefix from finalization signature verification to match kernel signing (both now use raw keccak256) |
  | — | Replace hardcoded `PartialDecryptionTimeoutBlocks` constant with on-chain `DecryptTimeout` param via `k.GetParams()` |

  ### KernelRouter Disconnect (`fix(dkg)` — `92a2ed59`)

  | Finding | Resolution |
  |---------|-----------|
  | CDR-078 | Clean up `ccByEP` mapping in `Disconnect()` so `disconnectedEndpoints()` recognizes the endpoint as available for reconnection |

  ### Default Parameters (`chore(dkg)` — `98dc2787`)

  | Parameter | Before | After |
  |-----------|--------|-------|
  | RegistrationPeriod | 86400 (1 day in seconds) | 34560 blocks (1 day) |
  | DealingPeriod | 86400 (1 day in seconds) | 34560 blocks (1 day) |
  | FinalizationPeriod | 86400 (1 day in seconds) | 34560 blocks (1 day) |
  | ActivePeriod | 1814400 (21 days in seconds) | 138240 blocks (4 days) |
  | MinReqRegisteredParticipants | 3 | 5 |
  | MinReqFinalizedParticipants | 3 | 5 |
  | OperationalThreshold | 667 (66.7%) | 500 (50%) |
  | DkgCommitteeRewardPortion | 10% | 5% |
  | DecryptTimeout | 10000 | 200 blocks |

  Update default DKG parameters assuming 2.5s block time.

issue: none


client/x/dkg/keeper/vote.go
```diff
@@ -137,6 +137,15 @@ func (k *Keeper) PrepareVotes(ctx context.Context, commit abci.ExtendedCommitInf
...
 }
 
+// aggregateVotes merges all vote extension payloads into a single Vote.
+// Deduplication is intentionally NOT performed here because
+// (1) the CL cannot validate DKG message authenticity — only the kernel can,
+// (2) first-seen dedup on unsigned fields lets an earlier-sorted validator
+//
+//	suppress honest messages by broadcasting colliding fake entries, and
+//
+// (3) VerifyVoteExtension already caps max items per VE, preventing DoS.
```
jdubpark, line 147: FWIW: When we dequeue to pack into votes, we already cap the number to 80 following this logic in `x/dkg/keeper/vote.go`:
```go
const (
	// maxVoteExtensionSize is the maximum allowed size of a raw vote extension in bytes (256 KB).
	maxVoteExtensionSize = 256 << 10

	// maxItemsPerVote is the maximum number of deals, responses, or justifications
	// allowed in a single vote extension. With 80 validators, a single validator
	// produces at most 79 deals and 79 responses per round.
	maxItemsPerVote = 80
)

func (k *Keeper) ExtendVote(_ sdk.Context, _ *abci.RequestExtendVote) (*abci.ResponseExtendVote, error) {
	dequeuedDeals := k.DequeueDeals(maxItemsPerVote)
	dequeuedResponses := k.DequeueResponses(maxItemsPerVote)
	dequeuedJustifications := k.DequeueJustifications(maxItemsPerVote)

	bz, err := proto.Marshal(&types.Vote{
		Deals:          dequeuedDeals,
		Responses:      dequeuedResponses,
		Justifications: dequeuedJustifications,
	})
	if err != nil {
		return nil, errors.Wrap(err, "marshal vote")
	}

	return &abci.ResponseExtendVote{
		VoteExtension: bz,
	}, nil
}
```

jdubpark (approved):
LGTM

Ramarti:
Running fix reviews...

Ramarti:
## CDR Fix Review: PR #755

**Methodology:** Fix-regression-auditor (FR-01..FR-06) + cross-repo verification with kernel `hans/audit-fix`
**Full report:** [`audit-agent/reports/CDR_FIX_REVIEW_PR755.md`](https://github.com/piplabs/story/tree/dkg/dev)

---

### Summary

| Category | Count |
|----------|-------|
| Fixes verified correct | 20 |
| Fixes with caveats | 3 |
| New issues found | 4 |
| CDR-069 (P1) still unfixed | 1 |

### CRITICAL: Cross-Repo Deployment Dependency

This PR removes the `\x19Ethereum Signed Message:\n32` prefix from finalization signature verification. Kernel `hans/audit-fix` (commit `2b14d75`) removes the same prefix from signing. **Both must be deployed simultaneously** -- deploying one without the other breaks all DKG finalization.

### CRITICAL: CDR-069 (P1) Remains Unfixed

`distributeCDRFee()` still distributes fees to ALL addresses in `CDRPartialSubmitCount` without committee membership check. Combined with CDR-057 (no access control on `submitEncryptedPartialDecryption`), fee pool drain is permissionlessly exploitable.

### CRITICAL: CDR-086 (P1) Pre-Existing Encoding Mismatch

CL `verifyPartialDecryptionSignature` uses **raw byte concatenation**. Kernel `signPartialDecryptResponse` uses **RLP encoding**. Different hashes -> every partial decryption signature fails verification. Threshold decryption is broken. Pre-existing since kernel commit `c36bcb3`; this PR's new comment on `dkg_handler.go:365` claiming "kernel and CL are consistent" is misleading.

---

### V5 Finding Status Update

| Finding | Sev | Status |
|---------|-----|--------|
| CDR-068 | P1 | **FIXED** (sort + CDR-017 overflow) |
| CDR-069 | P1 | UNFIXED |
| CDR-070 | P2 | **FIXED** (kernel `17b832c`) |
| CDR-071 | P2 | UNFIXED |
| CDR-072 | P2 | UNFIXED |
| CDR-073 | P2 | **PARTIAL** (RoundContext evicted; PIDCache left) |
| CDR-074 | P2 | UNFIXED |
| CDR-075 | P3 | UNFIXED |
| CDR-076 | P3 | UNFIXED |
| CDR-077 | P3 | UNFIXED |
| CDR-078 | P3 | **FIXED** (`92a2ed5`) |
| CDR-079-082 | P4-P5 | UNFIXED |

### V4 Findings Newly Fixed

CDR-039 (UUID overflow), CDR-050 (CDR nonReentrant), STOR-2 (new validator), STOR-7/10 (dedup), STOR-11 (finalize auth), STOR-14 (recipient filter), STOR-24 (truncation)

Ramarti:
@0xHansLee the new findings will be part of a different full report. IMO is OK not to fix them here. Will remove from comment

0xHansLee:
> @0xHansLee the new findings will be part of a different full report. IMO is OK not to fix them here. Will remove from comment

@Ramarti v5 will be the next version of the report?

client/x/dkg/keeper/vote.go
0xHansLee: An issue is reported that the dedup operations could erase earlier one, leading to malicious behavior. So removed it.

stevemilk:
CDR-069 is not valid since  all addresses in CDRPartialSubmitCount are already verified before updating the map.

0xHansLee:
> CRITICAL: Cross-Repo Deployment Dependency

Both will be deployed simultaneously.

> CRITICAL: CDR-069 (P1) Remains Unfixed

This is not a valid issue because `CDRPartialSubmitCount` is only for the members of DKG. 

> CRITICAL: CDR-086 (P1) Pre-Existing Encoding Mismatch

Will change to use RLP encoding in CL as well.

0xHansLee:
For the unfixed items, I moved the comments from each issue created on each repo:

- CDR-071: A specific binary of kernel is enforced to be used for all DKG members based on the MRENCLAVE, so the risk of this is quite low. We can just keep in mind that the `poly_persist.go` should be handled carefully when upgrading the kyber package. 
- CDR-072: While the observation is theoretically correct (a missing fsync between WriteFile and Rename could lead to data loss on power failure), the practical risk is extremely low: Narrow window — the failure would require a power loss in the exact window between WriteFile and Rename, which is extremely unlikely. We may revisit this as a low-priority hardening task in the future.
- CDR-074: There would not impact on the DKG and CDR operations even though the CL interacts with wrong enclave. It will rejected on contracts or CL. When upgrading the kernel, only the latest code commitment will be stored in current design, so we should not verify the code commitment based on the one stored on the contract.
- CDR-075: Mostly a maintainability concern, not a security issue. I think now we don't have to deal with this.
- CDR-076: Kernel router is only called in a separate go routine, so it should not impact on the liveness. It's invalid issue.
- CDR-077: Not valid. Encrypted deals can only be decrypted by the kernel during runtime.
- CDR-079: Not a valid issue in practice. AddCDRFeeToPool/RefundCDRFee are only called after the EVM event is decoded, and ProcessCDRFeeCollected/ProcessDKGPartialDecryptionSubmitted already return early when Amount Fee is nil.
- CDR-080: Not a valid issue. distributeCDRFee runs in BeginBlocker and only touches the previous pool, while DeliverTx processes EVM logs in order, so FeeCollected is applied before RefundCDRFee within the same tx.
- CDR-081: Round is always incremented by 1, so FIFO should be fine.
- CDR-082: Justification data is public one, so it should be fine.

0xHansLee:
> CRITICAL: CDR-086 (P1) Pre-Existing Encoding Mismatch

Fixed in [a4fa40f](https://github.com/piplabs/story/pull/755/commits/a4fa40f1ee362a34888c7a3cf3af836b6d1e314a)

stevemilk (approved):

PR #770, opened 2026-04-03
fix(evmengine): delete upgrade-info.json on cancelupgrade

### LINKED ISSUE #757: cancelUpgrade does not delete upgrade-info.json from disk
## Problem

When `cancelUpgrade()` is executed via TimelockController, the on-chain pending upgrade state is cleared (`ClearUpgradePlan` / `ResetPendingUpgrade`), but `data/upgrade-info.json` is not deleted from disk.

This file is written by `DumpUpgradeInfoToDisk` during `planUpgrade` processing. After cancel, it becomes stale.

## Impact

- **cosmovisor restart**: Reads stale `upgrade-info.json`, tries to switch to `upgrades/<name>/bin/story` which may not exist → crash loop
- **New binary startup**: `ReadUpgradeInfoFromDisk()` reads stale data, may trigger `StoreUpgrades.Added` at wrong height
- **Operator confusion**: Cancel appears successful (chain doesn't halt), but node restart fails unexpectedly

## Reproduction

1. `planUpgrade(name, H)` — `upgrade-info.json` written to disk
2. `cancelUpgrade()` — on-chain state cleared, **file remains**
3. Restart node — cosmovisor reads stale file → crash

## Suggested Fix

In `CancelUpgrade()` (`client/x/evmengine/keeper/upgrades.go`), after clearing on-chain state, delete the disk file:

```go
// Delete stale upgrade-info.json
infoPath, err := k.upgradeKeeper.GetUpgradeInfoPath()
if err == nil {
    if removeErr := os.Remove(infoPath); removeErr != nil && !os.IsNotExist(removeErr) {
        clog.Warn(ctx, "Failed to delete upgrade-info.json after cancel", removeErr)
    }
}
```

Requires adding `GetUpgradeInfoPath() (string, error)` to the `UpgradeKeeper` interface in `client/x/evmengine/types/expected_keepers.go`.

lucas2brh:
cancelUpgrade clears on-chain pending state but did not delete data/upgrade-info.json from disk, causing cosmovisor to still switch binaries at the originally planned height after cancel.

- Add GetUpgradeInfoPath to UpgradeKeeper interface
- Call os.Remove on upgrade-info.json in both isTerence and non-isTerence paths of CancelUpgrade
- Update mock and unit tests

Verified on devnet: planUpgrade then cancelUpgrade deletes upgrade-info.json on all 4 validators.

issue: closes #757

### DIFF
--- a/client/x/evmengine/keeper/upgrades.go
+++ b/client/x/evmengine/keeper/upgrades.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"encoding/hex"
 	"fmt"
+	"os"
 	"strconv"
 
 	upgradetypes "cosmossdk.io/x/upgrade/types"
@@ -193,14 +194,18 @@ func (k *Keeper) CancelUpgrade(ctx sdk.Context, isTerence bool) (err error) {
 		if err = k.ResetPendingUpgrade(ctx); err != nil {
 			return errors.Wrap(err, "failed to reset upgrade")
 		}
-
-		return nil
+	} else {
+		if err = k.upgradeKeeper.ClearUpgradePlan(ctx); errors.Is(err, sdkerrors.ErrInvalidRequest) {
+			return errors.WrapErrWithCode(errors.InvalidRequest, err)
+		} else if err != nil {
+			return errors.Wrap(err, "failed to clear upgrade plan")
+		}
 	}
 
-	if err = k.upgradeKeeper.ClearUpgradePlan(ctx); errors.Is(err, sdkerrors.ErrInvalidRequest) {
-		return errors.WrapErrWithCode(errors.InvalidRequest, err)
-	} else if err != nil {
-		return errors.Wrap(err, "failed to clear upgrade plan")
+	if infoPath, pathErr := k.upgradeKeeper.GetUpgradeInfoPath(); pathErr == nil {
+		if removeErr := os.Remove(infoPath); removeErr != nil && !os.IsNotExist(removeErr) {
+			clog.Warn(ctx, "Failed to delete upgrade-info.json after cancel", removeErr)
+		}
 	}
 
 	return nil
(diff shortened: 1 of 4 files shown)


client/x/evmengine/keeper/upgrades.go
```diff
@@ -193,14 +194,18 @@ func (k *Keeper) CancelUpgrade(ctx sdk.Context, isTerence bool) (err error) {
...
 	}
 
-	if err = k.upgradeKeeper.ClearUpgradePlan(ctx); errors.Is(err, sdkerrors.ErrInvalidRequest) {
-		return errors.WrapErrWithCode(errors.InvalidRequest, err)
-	} else if err != nil {
-		return errors.Wrap(err, "failed to clear upgrade plan")
+	if infoPath, pathErr := k.upgradeKeeper.GetUpgradeInfoPath(); pathErr == nil {
+		if removeErr := os.Remove(infoPath); removeErr != nil && !os.IsNotExist(removeErr) {
+			clog.Warn(ctx, "Failed to delete upgrade-info.json after cancel", removeErr)
+		}
```
0xHansLee, line 208: When GetUpgradeInfoPath() returns an error, the file deletion is silently skipped with no log. Consider adding a warning log so operators can tell if the path lookup failed:

```go
if infoPath, pathErr := k.upgradeKeeper.GetUpgradeInfoPath(); pathErr == nil {
    if removeErr := os.Remove(infoPath); removeErr != nil && !os.IsNotExist(removeErr) {
        clog.Warn(ctx, "Failed to delete upgrade-info.json after cancel", removeErr)
    }
} else {
    clog.Warn(ctx, "Failed to get upgrade-info.json path for cleanup", pathErr)
}
```

0xHansLee (approved):

ramtinms (approved):

PR #771, opened 2026-04-04
fix(dkg): use gasless context for nodes enabled dkg svc

0xHansLee:
### Problem

DKG-enabled (TEE) and DKG-disabled (non-TEE) validators produce **different `GasUsed`** for the same `MsgAddDkgVote` transaction. `GasUsed` is included in `deterministicExecTxResult` which feeds into `LastResultsHash`. Different `GasUsed` → different `LastResultsHash`.

**Root cause**: Inside `isDKGSvcEnabled` blocks, the keeper reads from the Cosmos KV store (e.g., `shouldProcessResponses()` → `getLatestActiveDKGNetwork()`). Story uses `SkipAnteHandler: true`, so the `InfiniteGasMeter` from `getContextForTx` persists for the entire tx. TEE nodes execute these KV reads and accumulate gas; non-TEE nodes skip them entirely — producing different `GasUsed` for the same block.

### Affected code paths (all inside `isDKGSvcEnabled` guards)

| Function | KV Read |
|----------|---------|
| `ProcessResponses` | `shouldProcessResponses()` → `getLatestActiveDKGNetwork()` |
| `BeginDealing` | `ensureSessionIndex()` → `getDKGRegistration()`, `shouldDeal()` → `getLatestActiveDKGNetwork()` |
| `InitiateDKGRound` | `isAlreadyRegistered()` → `getDKGRegistration()`, `getOldCodeCommitment()` |
| `BeginBlocker` → `ResumeDKGService` | All of the above via `resumeFailedSession()` |
| `ThresholdDecryptRequested` | `getDKGNetwork()` |

### Solution: Gasless Context

Introduce `gaslessSDKContext()` — a helper that returns a copy of the SDK context with a **fresh `InfiniteGasMeter`**. KV reads performed on this context succeed (same underlying multi-store) but do not contribute to the caller's `GasUsed`.

```go
func gaslessSDKContext(ctx context.Context) context.Context {
    sdkCtx := sdk.UnwrapSDKContext(ctx)
    return sdkCtx.WithGasMeter(storetypes.NewInfiniteGasMeter())
}
```

Each `isDKGSvcEnabled` block wraps its KV reads with `gaslessSDKContext(ctx)` before proceeding. The code structure remains nearly identical to the original — no function signature changes, no reads moved outside guards.

### Why not move reads outside `isDKGSvcEnabled`?

The fix approach moved KV reads **outside** the `isDKGSvcEnabled` guard so all nodes perform identical reads. While this works for new chains, it **breaks backward compatibility**:

### Backward compatibility

The gasless approach preserves identical `GasUsed` across all binary/config combinations:

| Node | Binary | DKG | GasUsed |
|------|--------|-----|---------|
| 1.6.1 disabled | old | No | `G_base` |
| 1.6.1 enabled | old | Yes | `G_base + N` (diverges — existing bug) |
| **1.6.2 disabled** | **new** | No | **`G_base`** |
| **1.6.2 enabled** | **new** | Yes | **`G_base`** (gasless reads) |

**Genesis sync**: A 1.6.2 node (DKG enabled or disabled) replaying from genesis produces `G_base` at every block — matching the chain state recorded by the DKG-disabled majority. No state mismatch.

**Rolling upgrade**: 1.6.1 disabled + 1.6.2 (any config) all produce `G_base` → 2/3+ consensus maintained throughout the upgrade. The small number of 1.6.1 enabled nodes that were already diverging simply upgrade to 1.6.2 to resolve.

**No hardfork required.**

issue: none

### DIFF
--- a/client/x/dkg/keeper/abci.go
+++ b/client/x/dkg/keeper/abci.go
@@ -65,8 +65,12 @@ func (k *Keeper) BeginBlocker(ctx context.Context) error {
 	}
 
 	if k.isDKGSvcEnabled {
+		// Use a gasless context for KV reads inside isDKGSvcEnabled so that
+		// DKG-enabled and DKG-disabled nodes produce identical GasUsed.
+		gaslessCtx := gaslessSDKContext(ctx)
+
 		// Resume stuck or failed DKG sessions every block.
-		k.ResumeDKGService(ctx, latestRound)
+		k.ResumeDKGService(gaslessCtx, latestRound)
 
 		// Retry cached deals/responses/justifications that failed kernel processing.
 		// Deals are replayed before responses (kyber requires deal-before-response order).
--- a/client/x/dkg/keeper/dkg_dealing.go
+++ b/client/x/dkg/keeper/dkg_dealing.go
@@ -55,10 +55,14 @@ func (k *Keeper) BeginDealing(ctx context.Context, latestRound *types.DKGNetwork
 	}
 
 	if k.isDKGSvcEnabled {
+		// Use a gasless context for KV reads inside isDKGSvcEnabled so that
+		// DKG-enabled and DKG-disabled nodes produce identical GasUsed.
+		gaslessCtx := gaslessSDKContext(ctx)
+
 		// Set session.Index from on-chain registration while SDK context is available.
 		// Session.Index stores the 1-based registration index used for deal/response
 		// routing (converted to 0-based for Kyber) and threshold decryption PID.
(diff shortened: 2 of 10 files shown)


stevemilk (approved):

stevemilk:
LGTM. One concern: If we don't plan to use the current approach for Mainnet — for example, if we change how non-DKG nodes behave — we should do another hard fork on Aeneid first to validate the change before rolling it out to Mainnet.

PR #775, opened 2026-04-06
fix(dkg): skip caching duplicate deals rejected by kernel

### LINKED ISSUE #776: DKG deal retry loop: duplicate deals cached indefinitely after dealing phase
## Description and context

  DKG-enabled nodes enter an infinite deal retry loop after the dealing phase completes. Due to CometBFT's vote extension lifecycle (`ExtendVote` runs before `FinalizeBlock`), the same deals arrive in two consecutive blocks. The first delivery is processed successfully, but the second is rejected by the kernel with `"all N submitted deals were rejected"`. The error handler caches these rejected deals for retry, causing an infinite loop that never resolves.

  ## Steps to reproduce

  1. Run a devnet with 3+ DKG-enabled validators
  2. Wait for a DKG round to enter the dealing phase
  3. Observe story logs after deals are processed

  ## Experienced behavior

  `"retry failed: all N submitted deals were rejected"` logs repeat indefinitely every ~2 seconds on all DKG-enabled validators, consuming unnecessary kernel gRPC calls.

  ## Expected behavior

  After deals are successfully processed, duplicate deliveries from consecutive vote extensions should be silently dropped without caching for retry.

  ## Solution recommendation

  **Short-term (story):** Detect the specific `"submitted deals were rejected"` error with `InvalidArgument` gRPC status code and drop the deals instead of caching. Only transient errors (`Unavailable`, `Internal`) should be cached for retry.

  **Long-term (story-kernel):** Change `ProcessDeals` to return an empty success response (with no responses) instead of an error when all submitted deals are rejected. This makes it consistent with how `ProcessResponses` and `ProcessJustifications` already handle duplicate/failed items, and eliminates the need for error message matching on the story side.

0xHansLee:
## Summary

  - Fix infinite retry loop where already-processed DKG deals were continuously re-sent to the kernel
  - Add detailed processing count logs for deals, responses, and justifications

  ## Problem

  Due to CometBFT's vote extension lifecycle (`ExtendVote` runs before `FinalizeBlock`), the same deals can arrive in two consecutive blocks. The first delivery is processed successfully, but the second delivery is rejected by the kernel with `"all N submitted deals were rejected"` (kyber rejects duplicate deals).

  Previously, `handleDKGProcessDeals` treated all errors uniformly — including this rejection — and cached them for retry via `reprocessPendingIncomingData`. Since retrying the same duplicate deals always produces the same rejection, this created an **infinite retry loop** that never resolves.

  Notably, responses don't have this issue because the kernel's `ProcessResponses` does not return an error for duplicates — it silently skips them and returns success.

  ## Fix

  Detect the specific `"submitted deals were rejected"` error message with `InvalidArgument` gRPC status code and drop the deals instead of caching. Other `InvalidArgument` errors (e.g., invalid round, empty deals) and transient errors (`Unavailable`, `Internal`) are still cached for retry as before.

issue: none

### DIFF
--- a/client/x/dkg/keeper/dkg_svc_dealing.go
+++ b/client/x/dkg/keeper/dkg_svc_dealing.go
@@ -5,10 +5,14 @@ import (
 	"context"
 	"encoding/hex"
 	"slices"
+	"strings"
 
 	"github.com/piplabs/story/client/x/dkg/types"
 	"github.com/piplabs/story/lib/errors"
 	"github.com/piplabs/story/lib/log"
+
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 )
 
 // handleDKGDealing handles the dealing phase event.
@@ -189,6 +193,17 @@ func (k *Keeper) handleDKGProcessDeals(ctx context.Context, dkgNetwork *types.DK
 
 		return nil
 	}); err != nil {
+		// "all N submitted deals were rejected" means every deal was already processed
+		// by kyber (duplicate delivery via consecutive vote extensions). Retrying the
+		// same data will always fail, so drop instead of caching.
+		if status.Code(err) == codes.InvalidArgument && strings.Contains(status.Convert(err).Message(), "submitted deals were rejected") {
+			log.Warn(ctx, "All deals already processed by kernel; dropping duplicates", err,
+				"round", dkgNetwork.Round,
+			)
+
+			return
+		}
+
 		// Cache unprocessed deals in memory for retry when kernel recovers.
 		// Not persisted to disk — process restart loses them (round will fail and retry).
 		cached := cachePendingIncomingDeals(deals, session.Index)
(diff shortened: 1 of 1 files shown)


stevemilk:
The `all %d submitted deals were rejected` error with `InvalidArgument` code doesn't exclusively indicate duplicate deals — it can be triggered by other error conditions as well. I assume the kernel-side error handling is intentionally left unchanged to avoid a kernel upgrade?

0xHansLee:
> The all %d submitted deals were rejected error with InvalidArgument code doesn't exclusively indicate duplicate deals — it can be triggered by other error conditions as well. I assume the kernel-side error handling is intentionally left unchanged to avoid a kernel upgrade?

Exactly. This could be included in the next release (v1.6.3). Also, I'm testing the fix on story-kernel side. The fix on story-kernel also be included in the next release(v0.1.1), but I guess it would not be applied in the near future.

stevemilk (approved):

barry-peng-story (approved):

PR #781, opened 2026-04-07
fix: drain and add back decrypt queue

jdubpark:
Fix a TOCTOU race in `processDecryptQueue` where the background decrypt worker's `GetDecryptRequests` → process → `SetDecryptRequests(remaining)` pattern silently overwrites requests added by the ABCI thread during the 3-6s processing window (kernel RPC + contract tx). This caused sequential same-wallet VaultRead requests to receive zero partial decryptions when the second read landed while the first was being processed.

Replace the get-process-set pattern with drain-and-addback: `DrainDecryptRequests` atomically swaps out the queue, the worker processes its private copy without holding locks, and failed requests are re-added via `AddDecryptRequest` — preserving any new requests the ABCI thread appended during processing.

### DIFF
--- a/client/x/dkg/keeper/dkg_svc.go
+++ b/client/x/dkg/keeper/dkg_svc.go
@@ -293,7 +293,9 @@ func (k *Keeper) processDecryptQueue(ctx context.Context) {
 
 	sessions := k.stateManager.ListSessions()
 	for _, session := range sessions {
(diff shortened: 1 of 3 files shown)


0xHansLee (approved):
LGTM

stevemilk (approved):

PR #783, opened 2026-04-08
fix(dkg): run threshold decrypt requested dkg service work asynchronously

### LINKED ISSUE #777: ThresholdDecryptRequested blocks FinalizeBlock on DKG-enabled nodes
## Description and context

  `ThresholdDecryptRequested` runs synchronously inside `FinalizeBlock`. On DKG-enabled nodes, after the early return guard (`!k.isDKGSvcEnabled`), the function performs KV reads (`getDKGNetwork`, `stateManager.GetSession`) and state writes (`stateManager.UpdateSession`) while still in the `FinalizeBlock` context.
   This blocks block finalization until all decrypt requests in the block are fully processed.

  For blocks containing many `ThresholdDecryptRequested` events, this adds latency to `FinalizeBlock` on DKG-enabled nodes compared to DKG-disabled nodes.

  ## Expected behavior

  The DKG-service-specific work (session lookup, `AddDecryptRequest`, `UpdateSession`) should run asynchronously via a goroutine, consistent with how `ProcessDeals`, `ProcessResponses`, and `handleDKGDealing` already handle their DKG-service work.

  ## Solution recommendation

  Run the DKG-service-specific work asynchronously in a goroutine to avoid blocking `FinalizeBlock`.

0xHansLee:
## Summary

  Move DKG-service-specific work in `ThresholdDecryptRequested` to an async goroutine to avoid blocking `FinalizeBlock` on DKG-enabled nodes.

  ## Problem

  `ThresholdDecryptRequested` runs synchronously inside `FinalizeBlock`. On DKG-enabled nodes, after passing the early return guards, it performs session lookup (`stateManager.GetSession`), appends the decrypt request (`AddDecryptRequest`), and persists the session (`UpdateSession`) — all while holding up block finalization.

  For blocks containing many `ThresholdDecryptRequested` events, this adds latency to `FinalizeBlock` on DKG-enabled nodes compared to DKG-disabled nodes.

  ## Fix

  Extract the DKG-service work into `handleThresholdDecryptRequest` and dispatch it via `dkgAsyncContext()` goroutine, following the same pattern used by `ProcessDeals`, `ProcessResponses`, and `BeginDealing`.

Close: #777

issue: #777

### DIFF
--- a/client/x/dkg/keeper/dkg_handler.go
+++ b/client/x/dkg/keeper/dkg_handler.go
@@ -475,38 +475,64 @@ func (k *Keeper) ThresholdDecryptRequested(ctx context.Context, round uint32, re
 		return nil
 	}
 
-	log.Info(ctx, "DKG ThresholdDecryptRequested event received",
-		"round", round,
-		"requester_pubkey_len", len(requesterPubKey),
-		"ciphertext_len", len(ciphertext),
-		"label_len", len(label),
-		"block_height", blockHeight,
-	)
-
-	session, err := k.stateManager.GetSession(round)
-	if err != nil {
-		return errors.Wrap(err, "failed to get DKG session for decrypt request")
-	}
-
-	// Record the request so the off-chain service can pick it up and produce a TDH2 partial decrypt.
-	session.AddDecryptRequest(types.DecryptRequest{
+	// Pre-compute values that require SDK context, then dispatch async.
+	decryptReq := types.DecryptRequest{
 		Round:           round,
 		Ciphertext:      ciphertext,
 		Label:           label[:],
 		RequesterPubKey: requesterPubKey,
 		Height:          blockHeight,
-	})
+	}
+
+	asyncCtx, cancel := dkgAsyncContext()
+
+	go func() {
+		defer cancel()
+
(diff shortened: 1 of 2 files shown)


client/x/dkg/keeper/dkg_handler.go
stevemilk, line 445: maybe better to start the new routine from here.

jdubpark (approved):

client/x/dkg/keeper/dkg_handler.go
0xHansLee: The KV store reads (getDKGNetwork, ActiveValSet check) can't go into the goroutine — they share the underlying multi-store reference, which isn't safe for concurrent reads during CometBFT's commit after FinalizeBlock returns.

This follows the same pattern as BeginDealing, ProcessResponses, etc.: pre-compute KV reads synchronously, pass results to the goroutine. These reads are simple KV lookups that take microseconds, so the blocking impact is negligible. The actual blocking work (session lookup, AddDecryptRequest, UpdateSession) is already in the goroutine.

stevemilk (approved):

PR #789, opened 2026-04-09
feat: generate partials for decrypt requests in parallel

stevemilk:
For #792 

Concurrency is now set to 10. May need to adjust the value according to metrics on kernel side later.
Here's the [unit test](https://github.com/piplabs/story-kernel/blob/feat/parallelProcessDecrypt/service/dkg_partial_decrypt_concurrency_test.go) on kernel side to verify if `kernelClient. PartialDecryptTDH2` can be parallelized. Need further tests on testnet though.

### DIFF
--- a/client/x/dkg/keeper/dkg_svc.go
+++ b/client/x/dkg/keeper/dkg_svc.go
@@ -12,6 +12,18 @@ import (
 	"github.com/piplabs/story/lib/log"
 )
 
+// decryptComputeResult holds the outcome of a single parallel kernel partial-decrypt call.
+type decryptComputeResult struct {
+	req  types.DecryptRequest
+	resp *types.PartialDecryptTDH2Response
+	err  error
+}
+
+// decryptChanBuf is the fixed buffer size for the parallel-decrypt result channel.
+// This caps in-flight results and slows producers when the serial submission
+// loop is slower than the parallel kernel calls.
+const decryptChanBuf = 10
+
 // dkgSvcRound tracks which DKG round is currently being processed by async
 // goroutines. Zero means no round is running. A higher round number always
 // supersedes a stale lock from a previous round, preventing cross-round
@@ -355,27 +367,7 @@ func (k *Keeper) processDecryptQueue(ctx context.Context) {
 			"pending_requests", len(requests),
 		)
 
-		for _, req := range requests {
-			if err := k.handleDecryptRequest(ctx, session, req); err != nil {
-				log.Error(ctx, "Failed to process decrypt request", err,
-					"session", session.GetSessionKey(),
-					"round", req.Round,
-					"ciphertext_len", len(req.Ciphertext),
-					"label_len", len(req.Label),
-					"requester_pub_key_len", len(req.RequesterPubKey),
-				)
-				// Re-add failed request; this appends to the live queue, preserving
-				// any new requests the ABCI thread added while we were processing.
-				session.AddDecryptRequest(req)
-
-				continue
-			}
-
-			log.Info(ctx, "Successfully processed decrypt request",
-				"session", session.GetSessionKey(),
-				"round", req.Round,
-			)
-		}
+		k.processDecryptRequests(ctx, session, requests)
 
 		if err := k.stateManager.UpdateSession(ctx, session); err != nil {
 			log.Error(ctx, "Failed to update session after processing decrypt queue", err,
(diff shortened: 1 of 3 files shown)


client/x/dkg/keeper/dkg_svc.go
```diff
@@ -385,24 +400,93 @@ func (k *Keeper) processDecryptQueue(ctx context.Context) {
...
+	// sends exactly one value via the deferred send (covers both normal return,
+	// panic via recover, and runtime.Goexit from test helpers).
+	for range requests {
+		r := <-resultCh
+		if r.err != nil {
+			log.Error(ctx, "Kernel partial decrypt failed", r.err,
+				"session", session.GetSessionKey(),
+				"round", r.req.Round,
+			)
+			session.AddDecryptRequest(r.req)
```
0xHansLee, line 450: Failed requests are re-queued and retried. If the error is permanent (e.g., malformed ciphertext, invalid label), this becomes an infinite retry loop.

We should add a retry counter to each `DecryptRequest` and drop it after a reasonable limit (e.g., 5–10 attempts) with a warning log. This is the same pattern we discussed for DKG deal/response/justification caching.

client/x/dkg/keeper/dkg_svc.go
```diff
@@ -385,24 +400,93 @@ func (k *Keeper) processDecryptQueue(ctx context.Context) {
...
+		go func(req types.DecryptRequest) {
+			// Pre-initialize with an error so that if runtime.Goexit() is called
+			// (e.g. t.Fatal from a test mock), the deferred send still fires and
+			// the consumer loop is not left blocked on an unreceived value.
+			// recover() inside computePartialDecrypt handles actual panics;
+			// this defer handles Goexit, which recover() cannot intercept.
+			result := decryptComputeResult{req: req, err: errors.New("goroutine exited without result")}
+			defer func() { resultCh <- result }()
+			result = k.computePartialDecrypt(ctx, session, req)
+		}(req)
```
0xHansLee: The current implementation spawns all goroutines immediately regardless of `decryptChanBuf`:

```go
bufSize := min(len(requests), decryptChanBuf) // channel buffer = 10
resultCh := make(chan decryptComputeResult, bufSize)

for _, req := range requests {
    go func(req types.DecryptRequest) { ... }(req) // ALL goroutines spawn at once
}
```

If 50 requests come in, 50 goroutines will call `PartialDecryptTDH2` concurrently. The channel buffer only limits how many completed results can be queued — it does not limit concurrent kernel RPCs.

If we want to cap concurrent kernel calls at `decryptChanBuf`, we need a semaphore. Also, the Phase 2 loop should handle context cancellation, and the drain must be timeout-protected in case a kernel RPC ignores cancellation like below:

```go
func (k *Keeper) processDecryptRequests(ctx context.Context, session *types.DKGSession, requests []types.DecryptRequest) {
  // Phase 1: parallel kernel calls with concurrency limit.
  resultCh := make(chan decryptComputeResult, len(requests))
  sem := make(chan struct{}, decryptChanBuf) // limits concurrent kernel RPCs

  for _, req := range requests {
    go func(req types.DecryptRequest) {
      sem <- struct{}{}        // acquire before kernel call
      defer func() { <-sem }() // release after

      result := decryptComputeResult{req: req, err: errors.New("goroutine exited without result")}
      defer func() { resultCh <- result }()
      result = k.computePartialDecrypt(ctx, session, req)
    }(req)
  }

  // Track which requests we've received results for.
  received := make(map[int]struct{})

  // Phase 2
  processed := 0
  for processed < len(requests) {
      select {
      case r := <-resultCh:
          received[r.idx] = struct{}{}
          processed++
          // ... handle result

      case <-ctx.Done():
          drainTimeout := time.After(30 * time.Second)
          for processed < len(requests) {
              select {
              case r := <-resultCh:
                  received[r.idx] = struct{}{}
                  processed++
                  session.AddDecryptRequest(r.req)
              case <-drainTimeout:
                  // Re-queue requests we never got results for.
                  // Goroutines won't leak — channel has room for all sends.
                  for i, req := range requests {
                      if _, ok := received[i]; !ok {
                          session.AddDecryptRequest(req)
                      }
                  }
                  return
              }
          }
          return
      }
  }
```

client/x/dkg/keeper/dkg_svc.go
stevemilk: Right. I planned to review the pattern you added and then we can reuse it here later.

client/x/dkg/keeper/dkg_svc.go
stevemilk: Good point. Added semaphore to control the concurrency.

stevemilk: The context cancelation issue seems not valid since the context is a process-lifetime one.

0xHansLee (approved):

PR #797, opened 2026-04-14
fix(server): cap query parameter nesting depth in buildmap to prevent oom

### LINKED ISSUE #796: Cap query parameter nesting depth in buildMap to prevent OOM
## Description and context

`buildMap()` in `client/server/utils/querymap.go` has O(D²) algorithmic complexity where D is the number of dot-separated segments in a query parameter key. A single HTTP request with a deeply nested key (e.g., a.a.a...a with 6000 segments, fitting within the 8KB MaxHeaderBytes limit) causes ~836 MB of heap allocation and takes ~1.5s to process. This affects all 21 endpoints using `AutoWrap`.

## Steps to reproduce

1. Start a story node with --api-enable
2. Send: curl "http://localhost:1317/staking/validators?$(python3 -c "print('a.'*6000+'a')")=x"
3. Observe RSS spike of ~836 MB for a single request

## Experienced behavior

Single request: 4ms → 1573ms, +836 MB RSS. 20 concurrent requests cause write timeout (30s). Higher concurrency can OOM the node.

## Expected behavior

Deeply nested query keys should be rejected or bounded so that a single request cannot allocate disproportionate memory.

## Solution recommendation

Cap recursion depth in `buildMap()`. All current request structs use max depth 2 (pagination.limit). A limit of 5 is sufficient.

```go
const maxQueryDepth = 5

func buildMap(query url.Values, prefix ...string) (ret map[string]any) {
    if len(prefix) > maxQueryDepth {
        return nil
    }
    // ... rest unchanged
}
```

0xHansLee:
## Summary

Cap recursion depth in `buildMap()` to prevent O(D²) memory amplification from deeply nested query parameter keys.

## Changes

- `querymap.go`: add `maxQueryDepth = 5` guard at the top of `buildMap()`. All current request structs use max depth 2 (`pagination.limit`), so 5 provides sufficient headroom.
- `querymap_test.go` (new): unit tests for normal usage, exact-at-limit, over-limit, and malicious depth. Benchmarks for normal vs deep-key scenarios.

Verified on a local single-validator network that a single malicious request (depth=6000) drops from 1,573 ms / +836 MB RSS to 1 ms / +0.4 MB RSS, while normal API requests remain unaffected.

Close: #796 

issue: #796

### DIFF
--- a/client/server/utils/querymap.go
+++ b/client/server/utils/querymap.go
@@ -100,7 +100,16 @@ func stringArrayToNative() mapstructure.DecodeHookFunc {
 	}
 }
 
+// maxQueryDepth limits the nesting depth of dot-separated query parameter keys
+// to prevent O(D^2) CPU and memory consumption from deeply nested keys.
+// Current max actual usage is depth 2 (e.g. "pagination.limit").
+const maxQueryDepth = 5
+
 func buildMap(query url.Values, prefix ...string) (ret map[string]any) {
+	if len(prefix) > maxQueryDepth {
+		return nil
+	}
+
 	fullPrefix := strings.Join(prefix, ".")
 	if len(fullPrefix) > 0 {
 		fullPrefix += "."
--- a/client/server/utils/querymap.go
+++ b/client/server/utils/querymap.go
@@ -2,6 +2,7 @@
 package utils
 
 import (
+	"fmt"
 	"math"
 	"math/big"
 	"net/url"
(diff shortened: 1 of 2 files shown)


stevemilk (approved):

client/server/utils/querymap.go
```diff
@@ -100,7 +100,16 @@ func stringArrayToNative() mapstructure.DecodeHookFunc {
 	}
 }
 
+// maxQueryDepth limits the nesting depth of dot-separated query parameter keys
+// to prevent O(D^2) CPU and memory consumption from deeply nested keys.
+// Current max actual usage is depth 2 (e.g. "pagination.limit").
+const maxQueryDepth = 5
+
 func buildMap(query url.Values, prefix ...string) (ret map[string]any) {
+	if len(prefix) > maxQueryDepth {
```
ramtinms: Do we need to return error or something, seems like a silent return

0xHansLee: Valid point. I changed to validate the depth in caller of `buildMap` function in [cc861b5](https://github.com/piplabs/story/pull/797/commits/cc861b567c9af5f3998e5e060e22dbcf60ca9d69)

ramtinms (approved):

PR #799, opened 2026-04-14
feat: batch submit encrypted partial decryptions

stevemilk:
## Summary
- Adds submitEncryptedPartialDecryptionBatch to the CDR contract, allowing validators to submit multiple partial decryptions in a single transaction (~40% gas saving per item at batch size 20)
- Rewrites processDecryptRequests in the CL keeper to compute partial decryptions in parallel and submit results in batches, replacing N individual contract calls with batched submissions
- Adds UpgradeCDR.s.sol upgrade script for the CDR v1.1.0 deployment

issue: #793


client/x/dkg/keeper/contract_client.go
```diff
@@ -288,6 +289,35 @@ func (c *ContractClient) SubmitEncryptedPartialDecryption(
...
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to pack submitEncryptedPartialDecryptionBatch call data")
+	}
+
+	baseFee, err := c.cdrContract.BaseFee(&bind.CallOpts{Context: ctx})
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to query CDR base fee for batch partial decryption")
+	}
+
+	totalFee := new(big.Int).Mul(baseFee, big.NewInt(int64(len(requests))))
```
0xHansLee: Why we need to multiply the length of the request? it's for calldata? I think we can just put basefee to `sendWithRetry`

client/x/dkg/keeper/dkg_svc.go
```diff
@@ -19,11 +20,17 @@ type decryptComputeResult struct {
...
+// decryptChanBuf limits the number of goroutines running kernel calls concurrently.
 const decryptChanBuf = 10
 
+// decryptBatchSize is the number of partial decryptions to submit in a single
+// batch contract call. The CDR contract enforces its own maxBatchSize;
+// this client-side value should be ≤ that limit.
+// One single submission tx consumes nearly 72k gas, including fixed overhead and per-decryption cost.
+// Batching reduces total gas by amortizing the fixed overhead across multiple decryptions.
+// With 20 decryptions, we get ~35% gas savings compared to single submissions, while keeping batch size manageable.
+const decryptBatchSize = 20
```
0xHansLee: Should we make this as a variable because this can be changed in the contract?

client/x/dkg/keeper/contract_client.go
```diff
@@ -288,6 +289,35 @@ func (c *ContractClient) SubmitEncryptedPartialDecryption(
 	})
 }
 
+// SubmitEncryptedPartialDecryptionBatch calls the batch submitEncryptedPartialDecryptionBatch
+// contract method, submitting multiple partial decryptions in a single transaction.
+func (c *ContractClient) SubmitEncryptedPartialDecryptionBatch(
```
0xHansLee, line 294: Can you add some unit tests for this func?

0xHansLee (commented):
Can you rerun `make contract-bindings`?

barry-peng-story (approved):

client/x/dkg/keeper/dkg_svc.go
stevemilk: Sounds good. Done

client/x/dkg/keeper/contract_client.go
stevemilk: Already added in cdr.t.sol

stevemilk:
> Can you rerun `make contract-bindings`?

done. there're some format adjustments in irrelevant files that we can ignore.

client/x/dkg/keeper/dkg_svc.go
0xHansLee: Oh, you added as a config. I meant adding to evmengine module to listen the event when the value is changed.

stevemilk: Contract defines the max batch size. Validators can adjust the batch they submitted according to their kernel perf and node spec. So i think we don't have to make these two values consistent

0xHansLee: Got you. Then, it should be less than the one stored in the contract?

client/config/dkg_config.go
```diff
@@ -48,6 +53,7 @@ func BindDKGFlags(flags *pflag.FlagSet, cfg *DKGConfig) {
 	flags.StringSliceVar(&cfg.KernelEndpoints, "dkg-kernel-endpoints", cfg.KernelEndpoints, "Comma-separated list of story-kernel (TEE) endpoints for DKG")
 	flags.StringVar(&cfg.EngineRPCEndpoint, "dkg-engine-rpc-endpoint", cfg.EngineRPCEndpoint, "The RPC endpoint of execution layer")
 	flags.Uint64Var(&cfg.EnclaveType, "dkg-enc-type", cfg.EnclaveType, "TEE enclave type identifier (e.g. 1 for SGX)")
+	flags.IntVar(&cfg.DecryptBatchSize, "dkg-decrypt-batch-size", cfg.DecryptBatchSize, "Number of partial decryptions per CDR batch call (must be ≤ contract maxBatchSize)")
```
0xHansLee, line 56: Regarding this, can you update `default.toml`, `story.toml`, and `config.toml.tmpl` under `story/client/config/`?

client/x/dkg/keeper/dkg_svc.go
stevemilk: right

client/config/dkg_config.go
stevemilk: updated

0xHansLee (approved):

PR #803, opened 2026-04-27
feat(dkg): expose GetAllDKGRegistrations through story-api

### LINKED ISSUE #802: [DKG] Add getAllRegistrationsByRound ABCI query for CDR SDK client-side signature verification
## Context

The CDR SDK (`piplabs/cdr-sdk`) has two modes for collecting partial decryptions from validators:
- `evm-events` — scans on-chain `EncryptedPartialDecryptionSubmitted` events and verifies the TEE signature on each partial locally against the validator's commPubKey
- `cosmos-abci` — queries the x/dkg keeper directly via CometBFT `abci_query` for faster polling

The `cosmos-abci` mode is currently broken for client-side use because the SDK cannot reliably obtain the correct commPubKey for the signing round. See [cdr-sdk#46](https://github.com/piplabs/cdr-sdk/issues/46) for the original bug report.

## Problem

To verify a partial's signature, the SDK needs the commPubKey that was active in the **round the validator signed under** (which is the currently *finalized* round, not the latest *registered* round — registration for round N+1 can happen while round N is still the active signing round).

The existing ABCI queries do not expose this reliably:

- `GetAllVerifiedDKGRegistrations(round)` only returns registrations whose status is `verified`. **Once a round enters the `finalized` phase, the keeper clears these entries and the query returns an empty array** — so there is no way to fetch commPubKeys for the currently active (finalized) round from the keeper.
- `GetLatestActiveDKGNetwork()` returns the active round number and threshold but not per-validator commPubKeys.

Combining the two queries does not work for the primary use case: "given the round a partial was signed under, fetch the commPubKeys of the registered validators for that round."

## Observed on Aeneid (2026-04-23)

| Round | `GetAllVerifiedDKGRegistrations` | Status |
|:---:|:---|:---|
| 1 | empty | finalized |
| 2 | empty | finalized |
| 3 | empty | finalized |
| 4 | empty | finalized |
| **5 (active, finalized)** | **empty** | **active — partials signed with round-5 commPubKeys** |
| 6 | 5 registrations | registering |

Result: cosmos-abci mode's `getRegisteredValidators()` returns an empty map → every partial fails signature verification → `collectPartials()` times out with `got 0/N`.

## Requested change

Add a new ABCI query `GetAllRegistrationsByRound(round)` that returns all registrations for a given round.

The returned entries should include at least:
- `validator_addr`
- `enclave_comm_key` (the commPubKey)
- `round`
- `enclave_type` / `code_commitment` (for optional filtering, same as existing queries)

Behaviorally, this query must return **all** registrations for the given round regardless of their verification or finalization status, so that downstream clients can verify partial signatures after a round has finalized.

## Consumer

The CDR SDK's `Consumer.collectPartialsFromCosmos` path will use this new query by:
1. Reading the round off each observed partial event
2. Calling `GetAllRegistrationsByRound(round)` (cached per round)
3. Looking up the commPubKey for each submitting validator and verifying the signature locally

This is the same verification pattern as `evm-events` mode, but sourced from keeper state instead of EVM event logs.

## References

- cdr-sdk issue: https://github.com/piplabs/cdr-sdk/issues/46
- Slack discussion: proj-cdr channel, 2026-04-24 thread starting at 15:06 CST
- Related evm-events fix (lookback window): https://github.com/piplabs/cdr-sdk/pull/54

0xHansLee:
Closes #802.

## Summary

- Wire the existing `GetAllDKGRegistrations` keeper query through the story-api at `GET /dkg/registrations`. The CDR SDK (`cosmos-abci` mode) needs commPubKeys for partials signed under an already-finalized round, but `GetVerifiedDKGRegistrations` returns empty in that case because the keeper transitions entries from `Verified` to `Finalized` at finalization. `GetAllDKGRegistrations` already returns every registration regardless of status, so
  no new RPC is added — it's just exposed.
- Document all DKG endpoints in `client/server/README.md` with response examples sampled from Aeneid.
- Deprecate `code_commitment_hex` on `QueryGetDKGNetworkRequest`, `QueryGetDKGRegistrationRequest`, `QueryGetAllDKGRegistrationsRequest`, and `QueryGetAllVerifiedDKGRegistrationsRequest`. The keeper has never read this field; mark it deprecated in the proto and stop forwarding it from the story-api (kept on the wire and in URL paths for backward compatibility).
- Fix `client/server/utils.QueryMapToVal`: the `stringArrayToNative` decode hook was missing cases for `uint32` (and other widths beyond `uint64`/`int`), so every DKG endpoint with a `uint32 round` query parameter returned HTTP 500 (`unconvertible type '[]string'`). Add the missing integer-width cases plus a unit test.

issue: #802


barry-peng-story (approved):

stevemilk (approved):

yingyangxu2026:
### SDK consumer note — `GetCDRPartials` is the missing piece for full REST migration

Evaluating this PR from the angle of moving the CDR SDK off the CometBFT `abci_query` path entirely and onto the Story-API REST. With this PR merged, the **registry-query side** is fully covered for SDK use:

- `/dkg/registrations?round=N` (new in this PR) — commPubKey lookup including finalized rounds (closes cdr-sdk#46)
- `/dkg/registrations/verified?round=N` — works again once the `uint32` fix in `stringArrayToNative` lands
- `/dkg/latest_active`, `/dkg/dkg_network?round=N`, `/dkg/global_public_key` — already working today

The remaining critical gap is the **partial-decryption data plane**. The SDK's `accessCDR` flow needs to fetch the encrypted partials a validator submitted for a given `(uuid, requesterPubKey)`. The keeper method `GetCDRPartials` already exists in `client/x/dkg/keeper/query.go` (it's what the SDK's `cosmos-abci` mode currently calls via `abci_query`), but there is no REST handler registered for it:

```
$ curl 'http://172.192.41.96:1317/dkg/cdr_partials?uuid=42&requester_pub_key_hex=00'
404 page not found
```

Without this endpoint, the SDK cannot complete a full REST-only `accessCDR` — partial collection has to fall back to either `eth_getLogs` (block scanning, the path we're trying to retire) or `abci_query` on port 26657 (the path we'd like to stop maintaining alongside REST).

The fix is the same shape as the `GetAllDKGRegistrations` wire-up in this PR — the gRPC method already exists, so it's just an HTTP route + request struct + README entry.

0xHansLee:
@yingyangxu2026 I will work on that for another one you requested in the following PR. It is out of scope of this PR.

PR #804, opened 2026-04-27
feat(dkg): expose GetCDRPartials through story-api

0xHansLee:
## Summary

Wires up a REST handler for the existing `GetCDRPartials` keeper query so clients can collect partial decryption submissions for a given `(uuid, requesterPubKey)` pair over REST. Closes the gap raised in [#803 (comment)](https://github.com/piplabs/story/pull/803#issuecomment-4323990727).

## Endpoint

GET /dkg/cdr_partials?uuid=&requester_pub_key_hex=

The server-wide envelope is `{"code": <int>, "msg": ..., "error": <string>}`. Success → HTTP 200. Any error (validation, keeper miss) → HTTP 500 with the message in `error`. REST-layer validation runs before the keeper:

- empty / `0x`-only pubkey → `requester_pub_key_hex is required`
- non-hex / odd-length → `requester_pub_key_hex must be valid hex`
- decoded length ≠ 65 → must be uncompressed secp256k1 (65 bytes)
- `uuid > math.MaxUint32-1` → `uuid out of range`

A leading `0x`/`0X` prefix is accepted (stripped before decoding).

issue: none

### DIFF
--- a/client/server/README.md
+++ b/client/server/README.md
@@ -1518,6 +1518,76 @@ Round 1 on Aeneid still has 5 entries in `Verified` status. `enclave_report` is 
 }
 ```
 
+## GetCDRPartials
+
+URL: [GET] /dkg/cdr_partials
+
+Returns the partial decryption submissions stored for a `(uuid, requesterPubKey)` pair, grouped by DKG round and ciphertext. Within each group the response carries the round's `threshold` and a precomputed `threshold_met` flag so callers do not need to query the network separately.
+
+`requester_pub_key_hex` is the hex encoding of the requester's public key as used when the decrypt request was submitted on-chain. A leading `0x` (or `0X`) prefix is accepted and stripped before decoding. Internally the keeper iterates the partial-decryption store under the `(requester_pub_key, label)` prefix where `label` is the 32-byte big-endian encoding of `uuid`.
+
+Bytes-typed fields (`encrypted_partial`, `ephemeral_pub_key`, `pub_share`, `label`, `ciphertext`) are encoded as base64 strings on the wire.
+
+### Query Params
+| Name                  | Type   | Example                                                       | Required |
+|-----------------------|--------|---------------------------------------------------------------|----------|
+| uuid                  | uint32 | 42                                                            |    ✔     |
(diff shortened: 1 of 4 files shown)


yingyangxu2026:
LGTM — clean PR, faster turnaround than I expected. The validation is a
  nice quality-bar bump over the existing handlers — `[0, MaxUint32-1]`
  uuid bound with an inline comment about the contract-side constraint is
  exactly the kind of detail that saves debugging time later.

  One question: the PR description says invalid inputs return HTTP 400,
  but the validate() function returns plain errors via errors.New /
  errors.Wrap without an explicit status mapping. Is `utils.AutoWrap`
  expected to convert those into 400, or do they currently surface as
  500? Worth either clarifying in the description or pinning behavior
  with an integration test.

  Optional: consider stripping `0x` prefix in the validator for
  ergonomics — the SDK strips it client-side already, but other REST
  consumers may not, and `hex.DecodeString` will reject `0x...` since
  `x` isn't a hex character. README already says "no `0x` prefix" so
  this is a soft preference; current strict behavior is also defensible.

  The README addition explaining the `label` byte layout is great —
  saves SDK implementers from reverse-engineering it.

stevemilk (approved):

0xHansLee:
Thanks for pointing it out @yingyangxu2026. I modified the description and changed to support stripping 0x prefix.

yingyangxu2026:
### Re: 500-for-all-errors

  Just saw the update — thanks for documenting the current behavior in the README and pointing at `client/server/utils/wrap.go`. That answers both my earlier `validate()` 400-vs-500 question and the `codes.NotFound` 404-vs-500 case I was about to flag, so this PR is
  good as-is from my side.

  Two follow-up notes for planning, no action needed on this PR:

  1. SDK consumers that poll `/dkg/cdr_partials` need to grep `error` for `"code = NotFound"` to distinguish "no data yet" from "real 5xx", which is fragile. Is the wrap.go gRPC→HTTP mapping tracked anywhere, or should we treat the substring workaround as long-lived?
  2. Thanks for the `0x`-prefix tolerance — addresses the ergonomic point from earlier review. 👍

yingyangxu2026 (approved):

0xHansLee:
@yingyangxu2026 It is not tracked at the moment, but I think that's the point that we need some more improvements on story-api. Let's make this later.

Issue #808, opened 2026-05-05
dkg nodes send duplicated register txs if txs are already in mempool but are not mined within 60s


stevemilk:
DKG nodes retry registration when failures occur—such as transactions being rejected—which is reasonable. However, in some cases, transactions remain pending in the mempool but are not included in blocks in time, causing DKG nodes to resend the same transactions.

Although the DKG module can handle duplicate transactions, node operators end up paying additional fees for these duplicates.

PR #810, opened 2026-05-06
fix(dkg): track pending txs in ContractClient to avoid nonce gaps on timeout

stevemilk:
issue: #808 

## Problem
When `waitForTransaction` times out (60s), the transaction may still be in the mempool — it was never dropped, just not yet mined. If the caller retries the operation on the next tick, `sendWithRetry` fetches a fresh nonce (N+1) and broadcasts a new transaction while the original (nonce N) is still pending. This creates a nonce gap: the new transaction at N+1 is stuck until N is resolved, stalling all subsequent transactions from that validator.

For CDR operations like `Register` and `Finalize`, a duplicate submission is also wasteful — the contract rejects the second call, but the gas is still spent.

## Solution

Add a `pendingTxs map[string]*types.Transaction` to `ContractClient`, keyed by operation + round (e.g. `"register_1"`).

The `sendTx` closure stores the tx hash **before** `waitForTransaction` is called, so the hash survives a timeout. On the next invocation of the same operation for the same round, `checkAndResumePendingTx` inspects the stored tx:

- **Still pending** (receipt == nil) → return error, abort. No new tx is sent.
- **Mined + success** → return the existing receipt directly.
- **Mined + failed** → clear the entry, proceed to send a fresh tx.

`sendWithRetry` itself is unchanged. The pending-tx check is an entry-point guard in `Register` and `Finalize`.

Also extends `dkgAsyncTimeout` from 1min to 2min with a comment explaining the budget: ~60s for TEE kernel operations + ~60s for `waitForTransaction`.

## Test Plan

- `TestCheckAndResumePendingTx_NoPendingTx` — no entry, returns (nil, nil)
- `TestCheckAndResumePendingTx_StillPending` — receipt nil, returns error, map not cleared
- `TestCheckAndResumePendingTx_MinedSuccess` — returns receipt, map cleared
- `TestCheckAndResumePendingTx_MinedFailed` — returns (nil, nil), map cleared
- `TestRegister_PendingTxStillPending` — aborts before sending, sendCalls == 0
- `TestRegister_PendingTxSucceeded` — reuses receipt, sendCalls == 0
- `TestRegister_PendingTxFailed_SendsNewTx` — sends fresh tx after on-chain failure
- `TestRegister_StoresPendingTxOnWaitTimeout` — confirms hash is preserved after context cancellation
- `TestFinalize_PendingTxStillPending` / `TestFinalize_PendingTxSucceeded` — same coverage for Finalize

### DIFF
--- a/client/x/dkg/keeper/contract_client.go
+++ b/client/x/dkg/keeper/contract_client.go
@@ -514,9 +514,13 @@ func (c *ContractClient) checkAndResumePendingTx(ctx context.Context, key string
 		return nil, nil
 	}
 
-	receipt, _ := c.ethClient.TransactionReceipt(ctx, tx.Hash())
+	receipt, err := c.ethClient.TransactionReceipt(ctx, tx.Hash())
 	if receipt == nil {
-		return nil, errors.New("previous tx still pending, skipping to avoid nonce gap",
+		if errors.Is(err, ethereum.NotFound) {
+			return nil, errors.New("previous tx still pending, skipping to avoid nonce gap",
+				"key", key, "tx_hash", tx.Hash().Hex())
+		}
+		return nil, errors.Wrap(err, "failed to check pending transaction receipt",
 			"key", key, "tx_hash", tx.Hash().Hex())
 	}
 
--- a/client/x/dkg/keeper/contract_client.go
+++ b/client/x/dkg/keeper/contract_client.go
@@ -522,6 +522,10 @@ func (c *ContractClient) checkAndResumePendingTx(ctx context.Context, key string
 
 	c.clearPendingTx(key)
 	if receipt.Status == types.ReceiptStatusSuccessful {
+		log.Info(ctx, key+" succeeded",
+			"tx_hash", tx.Hash().Hex(),
+			"gas_used", receipt.GasUsed)
+
 		return receipt, nil
 	}
 	return nil, nil
--- a/client/x/dkg/keeper/contract_client.go
+++ b/client/x/dkg/keeper/contract_client.go
@@ -6,6 +6,7 @@ import (
 	"encoding/hex"
 	"fmt"
 	"math/big"
+	"sync"
 	"time"
 
 	"github.com/ethereum/go-ethereum"
@@ -75,6 +76,13 @@ type ContractClient struct {
 	privateKey      *ecdsa.PrivateKey
 	fromAddress     common.Address
 	chainID         *big.Int
+
+	// pendingTxs tracks the last-sent tx for each operation key (e.g. "register_1").
+	// If waitForTransaction times out, the tx may still be in the mempool.
+	// The next invocation of the same operation checks this map first to avoid
+	// sending a new tx with a higher nonce while the previous one is still pending.
+	pendingTxMu sync.Mutex
+	pendingTxs  map[string]*types.Transaction
 }
 
 // ContractConfig holds configuration for contract interaction.
@@ -143,6 +151,7 @@ func NewContractClient(ctx context.Context, engineEndpoint string, engineChainID
 		privateKey:      privateKey,
 		fromAddress:     fromAddress,
 		chainID:         chainID,
+		pendingTxs:      make(map[string]*types.Transaction),
 	}
 
 	log.Info(ctx, "Created contract client",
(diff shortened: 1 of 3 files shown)


client/x/dkg/keeper/contract_client.go
```diff
@@ -453,3 +488,41 @@ func (c *ContractClient) sendWithRetry(
...
+//   - (nil, nil)      if no pending tx exists or it was mined but failed — caller should proceed normally
+func (c *ContractClient) checkAndResumePendingTx(ctx context.Context, key string) (*types.Receipt, error) {
+	c.pendingTxMu.Lock()
+	tx, ok := c.pendingTxs[key]
+	c.pendingTxMu.Unlock()
+	if !ok {
+		return nil, nil
+	}
+
+	receipt, _ := c.ethClient.TransactionReceipt(ctx, tx.Hash())
```
0xHansLee: Should we check error here as well?

client/x/dkg/keeper/contract_client.go
```diff
@@ -200,9 +209,22 @@ func (c *ContractClient) Register(ctx context.Context, round uint32, enclaveType
 
 	log.Info(ctx, "DKG registration fee queried", "fee_wei", fee.String())
 
-	return c.sendWithRetry(ctx, "Register", c.dkgContractAddr, callData, fee, func(auth *bind.TransactOpts) (*types.Transaction, error) {
-		return c.dkgContract.Register(auth, enclaveReport, enclaveInstanceData, startBlockHeightBig, startBlockHash32, []byte{})
+	pendingKey := fmt.Sprintf("register_%d", round)
+	if receipt, err := c.checkAndResumePendingTx(ctx, pendingKey); receipt != nil || err != nil {
```
0xHansLee, line 213: If the tx is pending, can we make this retry with higher gas fee?

client/x/dkg/keeper/contract_client.go
```diff
@@ -453,3 +488,41 @@ func (c *ContractClient) sendWithRetry(
...
+
+	receipt, _ := c.ethClient.TransactionReceipt(ctx, tx.Hash())
+	if receipt == nil {
+		return nil, errors.New("previous tx still pending, skipping to avoid nonce gap",
+			"key", key, "tx_hash", tx.Hash().Hex())
+	}
+
+	c.clearPendingTx(key)
+	if receipt.Status == types.ReceiptStatusSuccessful {
+		return receipt, nil
```
0xHansLee, line 533: How about adding a log here that the pending tx has been successfully mined with the pending key?

0xHansLee (commented):
How about we add this to `SubmitEncryptedPartialDecryption` and `SubmitEncryptedPartialDecryptionBatch`?

client/x/dkg/keeper/contract_client.go
stevemilk: SuggestGasPrice already returns the network's current recommended gas price, which is sufficient for inclusion under normal conditions. The only reason a tx ends up in pendingTxs is that waitForTransaction timed out after 60 s — the tx is still queued and will be mined. 
If we retry with higher gas price every time the gas price may increase very fast while it's just network stability problem.

client/x/dkg/keeper/contract_client.go
stevemilk: We only care about the receipt here. So I ignored the error intentionally.

stevemilk:
> How about we add this to `SubmitEncryptedPartialDecryption` and `SubmitEncryptedPartialDecryptionBatch`?

Unlike Register/Finalize which are "exactly once per round", batches can be sent multiple times per round with different request sets.

client/x/dkg/keeper/contract_client.go
stevemilk: added

client/x/dkg/keeper/contract_client.go
0xHansLee: If an error is returned here, the receipt might be nil but the returned error from here might be `previous tx still pending, skipping to avoid nonce gap`, right? I thought that it would be better to make this explicit.

stevemilk: Sounds good. Now return different error msg while the error is from like RPC failure.

0xHansLee (approved):

PR #819, opened 2026-05-12
feat(cdr):  prune onchain partials

stevemilk:
issue: #820

### Problem
DKGPartialDecrypt accumulates one entry per validator per CDR decrypt request and is never cleaned up, causing unbounded IAVL store growth that inflates the AppHash all validators sign.

### Solution
Keep only the two most recent successfully formed DKG rounds' data on-chain. A secondary round-keyed index (prefix 12, key: {round:010d}_{primaryKey}) enables O(deleted entries) range deletion without a full table scan.

Pruning triggers in BeginBlocker when a round transitions to Active. The previous active round is captured before FinalizeDKGRound updates LatestActiveRound, giving cutoff = prevActiveRound.Round - 1. Failed intermediate rounds are not counted — only successful rounds define the window.

Upgrade handler only backfills the secondary index for existing entries — no deletions. Pruning starts on the first FinalizeDKGRound after the upgrade, where all nodes prune at the same height and consensus is preserved.

### Next steps
Off-chain archival of pruned partial decrypt submissions to a local database, allowing individual nodes to configure their own retention policy independently of consensus state. This will enable historical CDR result queries without any on-chain storage cost.


client/app/upgrades/v_1_7_0/upgrades.go
```diff
@@ -0,0 +1,36 @@
...
+	configurator module.Configurator,
+	keepers *keepers.Keepers,
+) upgradetypes.UpgradeHandler {
+	return func(ctx context.Context, _ upgradetypes.Plan, vm module.VersionMap) (module.VersionMap, error) {
+		log.Info(ctx, "Start v1.7.0 upgrade — backfilling DKG partial decrypt round index")
+
+		newVM, err := mm.RunMigrations(ctx, configurator, vm)
+		if err != nil {
+			return vm, err
+		}
```
0xHansLee: There is no need to call this because it's for calling migration of each module when ConsensusVersion of module is upgraded. We are not using the ConsensusVersion for modules, so we don't need to call this.

client/x/dkg/keeper/abci.go
```diff
@@ -98,7 +98,31 @@ func (k *Keeper) BeginBlocker(ctx context.Context) error {
...
+				if err := k.pruneOldPartialDecryptions(ctx, cutoff); err != nil {
+					log.Error(ctx, "Failed to prune old partial decryptions", err,
+						"new_active_round", latestRound.Round,
+						"prev_active_round", prevActiveRound.Round,
+						"cutoff_round", cutoff,
+					)
+					// Non-fatal: pruning failure must not halt the chain.
+				}
+			}
+			return nil
```
0xHansLee: Here, getting the previous latest active round and pruning should be executed only after the upgrade is activated because the pruning is handling the KV store. Otherwise it would lead to app hash mismatch when syncing from the genesis, especially after introducing the DKG.

client/x/dkg/keeper/dkg_partial_decryption.go
```diff
@@ -71,6 +90,99 @@ func (k *Keeper) setPartialDecryptionSubmission(
 		return errors.Wrap(err, "set partial decryption submission")
 	}
 
+	// Write secondary round index entry alongside the primary. Value is empty;
+	// presence is sufficient for range-delete pruning.
+	indexKey := dkgPartialDecryptRoundIndexKey(round, key)
+	if err := k.DKGPartialDecryptRoundIndex.Set(ctx, indexKey, []byte{}); err != nil {
+		return errors.Wrap(err, "set partial decrypt round index")
+	}
```
0xHansLee: This should also be executed only after the upgrade is activated. Otherwise, this could lead to app hash mismatch.

client/app/upgrades/v_1_7_0/upgrades.go
stevemilk: removed

client/x/dkg/keeper/abci.go
stevemilk: set an index after the target upgrade to indicate pruning is activated

client/x/dkg/keeper/dkg_partial_decryption.go
stevemilk: done

client/x/dkg/keeper/abci.go
```diff
@@ -98,7 +98,37 @@ func (k *Keeper) BeginBlocker(ctx context.Context) error {
 		case types.DKGStageFinalization:
 			return k.BeginFinalization(ctx, latestRound)
 		case types.DKGStageActive:
-			return k.FinalizeDKGRound(ctx, latestRound)
+			// Capture the previous active round before FinalizeDKGRound updates LatestActiveRound.
+			// Only do this after the v1.9.0 upgrade has activated the secondary index; on nodes
+			// replaying from genesis the index does not exist yet and must not be consulted.
+			var prevActiveRound *types.DKGNetwork
+			if k.isPartialDecryptIndexActive(ctx) {
```
0xHansLee, line 105: Instead of storing as a state, can we use `IsV190` flag as we've done in the prev upgrades for consistency?

stevemilk: sounds good

0xHansLee (approved):

Issue #798, opened 2026-04-14
Add retry limits for cached kernel request reprocessing (deals/responses/justifications/decrypt)


0xHansLee:
## Description and context

When kernel requests (ProcessDeals, ProcessResponses, ProcessJustifications, DecryptRequest) fail, they are cached in-memory and retried automatically. Currently, there is **no per-item retry limit** — cached items are retried indefinitely until they either succeed, the DKG stage transitions (for deals/responses/justifications), or the request becomes stale past `DefaultDecryptTimeout` (for decrypt requests).

This can lead to:
- Excessive log noise from permanently failing items that will never succeed (e.g., malformed data, incompatible state)
- Wasted kernel RPC calls on requests that are fundamentally broken
- Potential resource pressure from repeated goroutine spawning for items that always fail

**Affected code paths:**
- `reprocessPendingIncomingData()` in `client/x/dkg/keeper/dkg_svc_dealing.go` — retries pending deals/responses/justifications every block with no per-item attempt counter
- `processDecryptQueue()` in `client/x/dkg/keeper/dkg_svc.go` — retries failed decrypt requests every 3 seconds with no per-item attempt counter
- Caching functions: `cachePendingIncomingDeals()`, `cachePendingIncomingResponses()`, `cachePendingIncomingJustifications()`, and `session.AddDecryptRequest()`

## Solution recommendation

**Client-side retry limit**

1. Add retry metadata to cached items: Wrap each cached item with a retry counter (retryCount int) using wrapper structs (pendingDeal, pendingResponse, pendingJustification, PendingDecryptRequest).
2. Define a max retry constant (e.g. maxReprocessAttempts = 3)
3. Increment and drop on failure: handleDKGProcessDeals/Responses/Justifications accept wrapper types directly. On kernel call failure, increment retryCount per item and drop items exceeding the limit. For decrypt requests, processDecryptRequests increments RetryCount on the PendingDecryptRequest wrapper when re-queuing failed items.
4. Log dropped items at Warn level with relevant context (round, retry count).

> Limitation: Kernel RPCs (ProcessDeals, ProcessResponses, ProcessJustification) are batch all-or-nothing — on failure, 
> ALL items in the batch get their retryCount incremented, even if only one item caused the failure. This is acceptable for > transient failures (kernel down/restarting) but suboptimal when a single bad item poisons the batch.

**Kernel per-item error reporting**

[...]

PR #828, opened 2026-05-20
feat(dkg): retry-limit with per-item kernel failure reporting

0xHansLee:
**Reopened this PR as it was merged unexpectedly before.**

## Summary
- Per-item retry counter for deals, responses, justifications, and threshold-decrypt requests. Items that repeatedly fail kernel processing are dropped after a fixed retry cap instead of being retried forever.
- When `ProcessXResponse.rejected_<type>` is populated, only the items matching the reported indexes are requeued; successful items are left alone. RPC errors continue to be treated as batch failures.
- `cachePending*` helpers now log and emit a `op="dropped"` metric when the queue is at capacity (previously a silent drop).

## Compatibility
Old kernels that leave `rejected_<type>` empty fall back to the legacy "RPC-success-means-done" behaviour, so the client works against both new and old kernels.

## ⚠️ Coordinated upgrade required
This PR must be deployed together with the matching kernel change in https://github.com/piplabs/story-kernel/pull/63. The new `rejected_<type>` fields rely on the kernel-side changes to be populated; without them, per-item retry never triggers and behaviour is identical to the previous release.

issue: #798


stevemilk (approved):

yingyangxu2026:
Reset devnet with current branch and conducted 1-hour stress test. No anomalies detected. https://github.com/piplabs/cdr-sdk/actions/runs/26270984008

PR #68, opened 2024-08-28
test(evmstaking/keeper): add test cases for staking queue

zsystm:
Increased test coverage from 0% to 84.6%
- **What's Covered**
  - Most application logics written on the Story side, excluding SDK.
- **What’s Not Covered (Edge Cases Related to the SDK)**
  - failed to get iterator
  - failed to unmarshal

issue: #64

PR #831, opened 2026-05-20
feat(dkg): add TDXValidationHook for TDX attestation

0xHansLee:
## Summary

On-chain validator for TDX V4/V5 attestation quotes used by the DKG flow. Mirrors `SGXValidationHook` in structure (`Ownable2Step + Pausable`) with TDX-specific quote field offsets.

## Identity model (v3 — RTMR3-bound binary, RTMR2-included platform)

Split along the operator / cloud boundary, where "operator-controlled" means *the running Go binary*:

- **Binary identity** (operator-controlled): `keccak256(RTMR3)`, whitelisted via `EnclaveTypeData.codeCommitment` (same path as SGX `MRENCLAVE`). RTMR3 is self-extended by the kernel exactly once at startup with `SHA-384(/proc/self/exe)` (see [story-kernel #68](https://github.com/piplabs/story-kernel/pull/68)), so its post-extend value `SHA-384(0x00..00 || SHA-384(elf))` is fully determined by the running ELF.
- **Platform identity** (cloud-managed): `keccak256(MRTD || RTMR0 || RTMR1 || RTMR2)`, approved via the hook's `approvePlatform` admin. RTMR2 moves into the platform half because it measures TD initrd + kernel cmdline — both boot-image properties decided at TD launch by the firmware/boot chain, never the userspace Go binary loaded from disk after boot.

Why not `keccak256(RTMR2)` for the binary half: on every direct-launch cloud TD measured, RTMR2 contains the TD initrd + cmdline digest. The story-kernel Go binary is loaded from the rootfs after boot and is never fed into RTMR2 by any link of the boot chain — so a `keccak256(RTMR2)` rule would commit to a boot-image artifact, not to the running ELF. The v3 split corrects this by claiming RTMR3 explicitly for the userspace binary half.

This decouples binary release governance from cloud firmware vintage rotation, so each axis updates independently. `REPORT_DATA[0:32]` carries the kernel-bound caller commitment from `DKG.register` and is compared against `expectedDataCommitment`, mirroring SGX's instance binding.

## Onboarding flow

A TDX enclave type is enabled in two steps:

1. Governance: `DKG.whitelistEnclaveType(enclaveType, { codeCommitment = keccak256(RTMR3), validationHookAddr = tdxProxy })`.
2. Hook owner: `TDXValidationHook.approvePlatform(keccak256(MRTD || RTMR0 || RTMR1 || RTMR2), label)`.

## Contract changes

- `TDXValidationHook` (new). `_computeBinaryCommitment` reads `RTMR3` at offset 520; `_computePlatformCommitment` assembles the 192-byte preimage from MRTD (offset 184) + the contiguous RTMR0/RTMR1/RTMR2 block (offsets 376..519).
- `ITDXValidationHook` (new). Admin surface: `setAutomataValidationAddr`, `approvePlatform`, `revokePlatform`, `isPlatformApproved`.
- `IAutomataDcapAttestationFee`: adds 1-arg overload (PR #816 pattern) so standard TCB transitions take effect automatically.

`DKG.sol` / `IDKG.sol` are unchanged — ABI/storage compatible with the live deployment.

## Adversary cost — both halves must succeed

To pass `register()` an attacker must produce a TDX quote where:

- `RTMR3` hashes to an approved binary commitment, **and**
- `MRTD || RTMR0 || RTMR1 || RTMR2` hashes to an approved platform commitment.

The RTMR3 self-extend can only push values *into* RTMR3 (RTMRs are append-only). An attacker binary `B'` that self-extends RTMR3 with `SHA-384(honest_B)` to forge the binary commitment still has to be loaded by the boot chain — which measures the kernel image + initrd into RTMR1 and the initrd + cmdline into RTMR2. Either modification shifts the platform commitment off any honest entry, and the platform-approval check fails.


contracts/script/GenerateAlloc.s.sol
```diff
@@ -65,6 +66,41 @@ contract GenerateAlloc is Script {
...
+    bytes32 private constant TDX_PLATFORM_COMMITMENT_GCP_C3S4_V1_PLACEHOLDER =
+        keccak256("TDX_PLATFORM_COMMITMENT_GCP_C3S4_V1_PLACEHOLDER_v3");
+    bytes32 private constant TDX_PLATFORM_COMMITMENT_GCP_C3S4_V2_PLACEHOLDER =
+        keccak256("TDX_PLATFORM_COMMITMENT_GCP_C3S4_V2_PLACEHOLDER_v3");
+
+    // When DKG_INCLUDE_TDX is true the script deploys TDXValidationHook and whitelists
+    // enclaveType=2 with the binary commitment above, then calls approvePlatform for each
+    // configured platform commitment. Operators flip this to true on TDX-capable devnets;
+    // it is false by default so mainnet/SGX-only deployments are unaffected.
+    bool private constant DKG_INCLUDE_TDX = true;
```
stevemilk: should be false here?

stevemilk (approved):

ramtinms (approved):

contracts/script/GenerateAlloc.s.sol
0xHansLee: Removed the constant and make the script set TDXValidationHook contract by default in [ddf3d7d](https://github.com/piplabs/story/pull/831/commits/ddf3d7d66fc378c248f61b57c98626926f979a89)

Issue #826, opened 2026-05-19
DKG committee change: first new-committee dealing round stuck at stage=5 + active round's decrypt requests dropped on next-next round creation → TDH2 partial-decrypt outage during round N+2 pre-active window


yingyangxu2026:
## TL;DR

Any DKG committee change (resharing with new members, or any committee size adjustment) simultaneously triggers two compounding bugs:

1. **Bug #1 (story-kernel)** — The kernel reads chain state through a verified light client. When the `BeginDealing` block has just committed but the kernel's light client cache hasn't caught up, [`fetchRoundContext`](https://github.com/piplabs/story-kernel/blob/63e74a025e86919204d87aae24d777be2813dda7/service/round_context.go#L79) reads `Total=0` while registrations already number N. [`validateRegistrations`](https://github.com/piplabs/story-kernel/blob/63e74a025e86919204d87aae24d777be2813dda7/service/round_context.go#L116) returns an error immediately, **short-circuiting the `Threshold=0` retry loop that was explicitly designed for this light-client lag**. Early deals/responses are rejected and never re-broadcast. `ThresholdCertified()` never returns true. **Round N+1 is permanently stuck at stage=5.**

2. **Bug #2 (story)** — [`processDecryptQueue`](https://github.com/piplabs/story/blob/d693b064d39f8d96f44f445f671b9d21096468c7/client/x/dkg/keeper/dkg_svc.go#L308) uses a `maxRound - 2` heuristic to identify stale sessions. **The moment round N+2's session is created in chain state**, StateManager's `maxRound` jumps to N+2 and the condition `session.Round=N <= maxRound-2` fires, causing **the partial-decrypt requests for the currently active round N (which SDK clients are still legitimately using) to be drained from the queue every tick**.

**Compound effect:**

- During round N+1's reg / dealing / finalization phases: `maxRound` is still N+1, session N is NOT drained, **round N continues to serve normally**.
- **From the moment round N+2's session is created** (start of round N+2 registration phase): SDK clients send TDH2 read requests using round N's GPK from `/dkg/latest_active` → keeper enqueues them on session N → next `processDecryptQueue` tick drains them → kernel never receives them → SDK times out at 300s with `got 0/0 partials`.
- When round N+2 enters active stage: `latest_active` switches to N+2 → new reads use N+2's GPK → service recovers.

**Outage window = `round N+2 reg_period + dealing_period + finalization_period`.** Measured on devnet: all 1000 concurrent reads fail during this window.

[...]

PR #834, opened 2026-05-21
fix(dkg): preserve active round decrypt queue when successor session is created

0xHansLee:
## Summary

`processDecryptQueue` previously anchored its stale-drain heuristic on `max(s.Round)` across all in-memory sessions. When round `N+1` got stuck and round `N+2`'s session was created, that anchor jumped to `N+2` and unconditionally drained the still-serving round `N`'s decrypt queue.

This change anchors the drain on the latest round with `IsFinalized=true` (set by `handleDKGComplete` when the round transitions to `PhaseCompleted`), so a stuck or in-progress successor cannot drain the active round's queue.

issue: #826

### DIFF
--- a/client/x/dkg/keeper/dkg_svc.go
+++ b/client/x/dkg/keeper/dkg_svc.go
@@ -325,10 +325,11 @@ func (k *Keeper) processDecryptQueue(ctx context.Context) {
 
 	sessions := k.stateManager.ListSessions()
 
-	var maxRound uint32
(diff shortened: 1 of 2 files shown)


yingyangxu2026:
Reset devnet with current branch and conducted 1-hour stress test. No anomalies detected. https://github.com/piplabs/cdr-sdk/actions/runs/26279495155

Issue #833, opened 2026-05-21
[BUG] DKG resharing: dealers with failed or delayed dealing remain in committee, silently producing partial decryptions inconsistent with consensus polynomial


yingyangxu2026:
## TL;DR

During DKG resharing in round N's dealing phase, if a dealer:

- **(Case A)** fails to produce any deal at all (due to a kernel-internal error such as `rebuildResharingNextDKG: replay failed`), or
- **(Case B)** completes its dealing significantly later than peers, causing its local dealing-phase state to diverge from consensus

its local share will not lie on the consensus polynomial. But the chain currently **keeps Finalizing it**, leaves it in the committee permanently, and **accepts all its subsequent partial decryption submissions**.

The chain has **no mechanism** to detect this corruption — the validator's status remains `DKGRegStatusFinalized`, indistinguishable from healthy validators in the chain's view. The only signal is downstream: SDK `tdh2_combine` fails with `-16515071` (E_CRYPTO) due to mathematical inconsistency.

## Impact

- **CDR access failure rate**: depends on number of corrupted vals × probability SDK's threshold-subset selection picks one. Devnet round 80 had 2 corrupted vals (n=17, threshold=9); 1000-wallet perf test got **999/1000 failures**; single cli access reproduces 100% of the time.
- **Data availability**: GPK remains stable, no data loss. But all CDR access fails during the corrupted round unless caller manually identifies and excludes bad vals.
- **Self-healing window**: next round's resharing re-derives shares from scratch, so corrupted vals usually heal. Empirically confirmed: round 80 corruption → round 81 fully recovered. But the corrupted-round duration (~58 min on current devnet config) is a complete CDR outage from the user's perspective.
- **Operational blind spot**: no on-chain metric or event surfaces the corruption. Operators must observe downstream SDK combine failures to discover it.

---

## Reproduction

Anyone with the chain REST partial-decryption data plus a fixed requester keypair can reproducibly trigger `wasm_tdh2_combine -16515071` and pinpoint the exact corrupted validators.

### Steps

1. Fetch the full partial-decryption submissions for round 80 uuid=12236 (chain still serves this; not yet pruned):

```
[...]

PR #837, opened 2026-06-01
fix(dkg): invalidate finalized validators whose share is off the consensus polynomial

0xHansLee:
## Issue

A validator whose dealing-phase view diverged from consensus can finalize with a public key share that is not on the consensus polynomial, producing partial decryptions inconsistent with the committee.

## Fix
`FinalizeDKGRound` now verifies each finalized validator's public key share against the consensus public coefficients (`vss.VerifyPublicKeyShare`) and invalidates mismatches before counting the committee; the round is skipped if no consensus coefficients were set. 

issue: #833

### DIFF
--- a/client/x/dkg/keeper/dkg_finalization.go
+++ b/client/x/dkg/keeper/dkg_finalization.go
@@ -3,6 +3,10 @@ package keeper
 import (
 	"context"
 
+	"cosmossdk.io/collections"
+
+	"github.com/ethereum/go-ethereum/common"
+
 	"github.com/piplabs/story/client/x/dkg/types"
 	"github.com/piplabs/story/lib/errors"
 	"github.com/piplabs/story/lib/log"
@@ -27,14 +31,22 @@ func (k *Keeper) BeginFinalization(ctx context.Context, latestRound *types.DKGNe
 }
 
 func (k *Keeper) FinalizeDKGRound(ctx context.Context, latestRound *types.DKGNetwork) error {
-	// Guard: global public key must be set before finalizing. If it's empty,
-	// the round did not complete key generation and should not be activated.
-	if len(latestRound.GlobalPublicKey) == 0 {
-		log.Info(ctx, "Global public key not set, skipping to next round",
+	// Guard: consensus key material must be set before finalizing. If it's empty, the
+	// round did not complete key generation and should not be activated.
+	if len(latestRound.GlobalPublicKey) == 0 || len(latestRound.PublicCoeffs) == 0 {
+		log.Info(ctx, "Consensus key material not set, skipping to next round",
 			"round", latestRound.Round,
 		)
 
 		return k.SkipToNextRound(ctx, latestRound)
+	}
+
+	// Drop finalized validators whose finalize vote diverged from the consensus
+	// polynomial before counting.
+	if k.isV190Round(ctx, latestRound) {
+		if err := k.invalidateNonConsensusFinalizations(ctx, latestRound); err != nil {
+			return errors.Wrap(err, "failed to invalidate non-consensus finalizations")
+		}
 	}
 
 	finalizedCount, err := k.countDKGRegistrationsByStatus(ctx, latestRound.Round, types.DKGRegStatusFinalized)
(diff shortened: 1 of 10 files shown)


stevemilk (approved):

ramtinms (approved):

yingyangxu2026:
Thanks for putting this together — validating finalized shares against the consensus polynomial is a really useful guard. While regression-testing it on devnet I ran into a blocker I wanted to flag.

## Symptom
On a V190-active chain, `FinalizeDKGRound` → `invalidateDivergentShares` (`dkg_finalization.go:148`) ends up invalidating **every** finalized validator, so the round never reaches active and DKG can't bootstrap — each round dies at stage 5 and the chain rolls into the next one indefinitely.

## Root cause: the edwards25519 suite is applied to the GPK-layer points
`invalidateDivergentShares` builds the suite as `edwards25519.NewBlakeSHA256Ed25519()` (`dkg_finalization.go:148`) and passes it to `vss.VerifyPublicKeyShare` (`:154`) to unmarshal `reg.PubKeyShare` and `latestRound.PublicCoeffs`. Those fields don't appear to be edwards25519 points:

1. **VSS dealing layer** — the per-dealer polynomial commitments verified during dealing/justification are edwards25519, and the existing `verifyJustificationVSS` / `VerifyPedersenVSS` (`dkg_justification.go:25`) correctly use the ed25519 suite for them.
2. **Threshold-key (GPK) layer** — `GlobalPublicKey` / `PublicCoeffs` (`DKGNetwork` field 10) / `PubKeyShare` (`DKGRegistration` field 6) are the aggregated threshold public-key material, which seem to live on a different group than the ed25519 dealing commitments. (I haven't pinned the exact curve from the code — you'd know best — but it's demonstrably not ed25519, see below.)

`VerifyPublicKeyShare` operates on the layer-2 fields but receives the layer-1 ed25519 suite, so `verify.go:111` `UnmarshalBinary` returns `invalid Ed25519 curve point`. Since `invalidateDivergentShares` treats any verify error as "divergent", it invalidates the validator — including the healthy finalized ones.

## Evidence (devnet, LocalChainID `story-1001511`, V190 active, periods reg50/deal80/fin80, no snapshot)
Round 4 actually succeeded (GPK generated, 2 validators logged `DKG successfully finalized`), then got wiped out:

```
WARN Invalidated finalized validator: share not on consensus polynomial
  round=4 index=2 err="unmarshal public key share: invalid Ed25519 curve point"   (verify.go:111)
WARN Invalidated finalized validator: share not on consensus polynomial
  round=4 index=1 err="unmarshal public key share: invalid Ed25519 curve point"
```

A kyber edwards25519 point is always exactly 32 bytes; the on-chain values (base64-decoded) aren't:

| field | bytes |
|---|---|
| GlobalPublicKey | 31 |
| PublicCoeffs[*] | 30 / 31 |
| PubKeyShare (index=1) | 32 (still fails `UnmarshalBinary` → not a valid ed25519 point) |
| PubKeyShare (index=2) | 33 |
| PubKeyShare (index=3) | 2 (this validator genuinely never dealt; correctly invalidated by the missing-dealer path) |

Net: index 1/2 (legitimately finalized) are invalidated here, plus index 3 (truly missing) by the dealer check → all 3 invalidated → round 4 can't activate → repeats every round → DKG never bootstraps. This reproduces both with and without the post-setup S3 snapshot.

## Suggestion
`VerifyPublicKeyShare` should use the suite for the group that `GlobalPublicKey` / `PublicCoeffs` / `PubKeyShare` actually live on, rather than edwards25519. It might also be worth confirming whether those points are compatible with kyber's `share.PubPoly.Eval` at all — they're variable-length (30–33 bytes), which doesn't match kyber's fixed-size point encoding, so the verification approach itself may need a small rework.

## Repro
Fresh V190-active chain (LocalChainID `story-1001511`), reg50/deal80/fin80. After bringing it up, DKG never reaches active and rounds stay at stage 5. `journalctl -u story | grep -E "Invalidated|invalid Ed25519"` shows the invalidations above.

Happy to share full node logs or jump on a call if that helps. Thanks again!

PR #838, opened 2026-06-01
fix(dkg): invalidate dealers that submit no deal during the dealing phase

0xHansLee:
## Issue

A dealer that produces no deal at all is not caught by the VSS complaint path, yet its share is absent from the consensus polynomial.

## Fix
Dealers that submit a deal are recorded while processing vote extensions (`ProcessDeals`); verified dealers with no recorded deal are invalidated at `BeginFinalization`, and the dealt set is pruned once consumed. 

issue: #833

### DIFF
--- a/client/x/dkg/keeper/dkg_finalization.go
+++ b/client/x/dkg/keeper/dkg_finalization.go
@@ -13,6 +13,13 @@ import (
 )
 
 func (k *Keeper) BeginFinalization(ctx context.Context, latestRound *types.DKGNetwork) error {
+	// Invalidate dealers that never submitted a deal during the dealing phase.
+	if k.isV190Round(ctx, latestRound) {
+		if err := k.invalidateMissingDealers(ctx, latestRound); err != nil {
+			return errors.Wrap(err, "failed to invalidate missing dealers")
+		}
+	}
+
 	if err := k.emitBeginDKGFinalization(ctx, latestRound); err != nil {
 		return errors.Wrap(err, "failed to emit begin DKG finalization event")
 	}
@@ -196,4 +203,99 @@ func (k *Keeper) invalidateNonConsensusFinalizations(ctx context.Context, latest
 	}
 
 	return nil
+}
+
+// invalidateMissingDealers invalidates dealers that never submitted a deal during the
+// dealing phase. The expected dealer set is the committee that should have produced deals:
+// the previous active committee for resharing rounds, the current round otherwise. A missing
+// dealer is penalized only if it also holds a verified registration in the current round; a
+// dealer leaving the committee has no current registration and nothing to invalidate. The
+// dealt set recorded in ProcessDeals is pruned here.
+func (k *Keeper) invalidateMissingDealers(ctx context.Context, latestRound *types.DKGNetwork) error {
+	expectedDealers, err := k.expectedDealers(ctx, latestRound)
+	if err != nil {
+		return err
+	}
+
+	for _, dealer := range expectedDealers {
+		dealerAddr := common.HexToAddress(dealer)
+		key := dealtDealerKey(latestRound.Round, dealer)
+
+		dealt, err := k.DealtDealers.Has(ctx, key)
+		if err != nil {
+			return errors.Wrap(err, "failed to check dealt dealer", "dealer", dealer)
+		}
+
+		if dealt {
+			if err := k.DealtDealers.Remove(ctx, key); err != nil {
+				return errors.Wrap(err, "failed to prune dealt dealer", "dealer", dealer)
+			}
+
+			continue
+		}
+
+		// No deal from this dealer. Only penalize if it is also a verified member of the
+		// current round; a leaving member has no current registration to invalidate.
+		has, err := k.hasDKGRegistration(ctx, latestRound.Round, dealerAddr)
+		if err != nil {
+			return errors.Wrap(err, "failed to check current registration", "dealer", dealer)
+		}
+		if !has {
+			continue
+		}
+
+		reg, err := k.getDKGRegistration(ctx, latestRound.Round, dealerAddr)
+		if err != nil {
+			return errors.Wrap(err, "failed to get current registration", "dealer", dealer)
+		}
+		if reg.Status != types.DKGRegStatusVerified {
+			continue
+		}
+
+		reg.Status = types.DKGRegStatusInvalidated
+		if err := k.setDKGRegistration(ctx, dealerAddr, reg); err != nil {
+			return errors.Wrap(err, "failed to invalidate missing dealer",
+				"validator", dealer,
+				"index", reg.Index,
+			)
+		}
+
+		log.Warn(ctx, "Invalidated dealer: no deal submitted during dealing phase",
+			nil,
+			"round", latestRound.Round,
+			"validator", dealer,
+			"index", reg.Index,
+		)
+	}
+
+	return nil
+}
+
+// expectedDealers returns the validator addresses of the committee that should produce deals
+// in latestRound: the previous active committee for resharing rounds, the current round's
+// registrations otherwise.
+func (k *Keeper) expectedDealers(ctx context.Context, latestRound *types.DKGNetwork) ([]string, error) {
+	if latestRound.IsResharing {
+		prevActive, err := k.getLatestActiveDKGNetwork(ctx)
+		if err != nil {
+			return nil, errors.Wrap(err, "failed to get previous active round")
+		}
+		if prevActive == nil {
+			return nil, nil
+		}
+
+		return prevActive.ActiveValSet, nil
+	}
+
+	regs, err := k.getDKGRegistrationsByRound(ctx, latestRound.Round)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to fetch DKG registrations")
+	}
+
+	dealers := make([]string, len(regs))
+	for i := range regs {
+		dealers[i] = regs[i].ValidatorAddr
+	}
+
+	return dealers, nil
 }
--- a/client/x/dkg/keeper/dkg_dealing.go
+++ b/client/x/dkg/keeper/dkg_dealing.go
@@ -184,6 +184,14 @@ func (k *Keeper) ProcessJustifications(ctx context.Context, latestRound *types.D
 }
 
 func (k *Keeper) ProcessDeals(ctx context.Context, latestRound *types.DKGNetwork, deals []types.Deal) error {
+	// Record which dealers submitted a deal so missing dealers can be invalidated at
+	// BeginFinalization.
+	if k.isV190Round(ctx, latestRound) {
+		if err := k.markDealersDealt(ctx, latestRound, deals); err != nil {
+			return errors.Wrap(err, "failed to mark dealers dealt")
+		}
+	}
+
 	if err := k.emitBeginProcessDeals(ctx, latestRound, deals); err != nil {
 		return errors.Wrap(err, "failed to emit begin process deals event")
 	}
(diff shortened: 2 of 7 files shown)


client/x/dkg/keeper/dkg_finalization.go
```diff
@@ -14,6 +14,13 @@ import (
 )
 
 func (k *Keeper) BeginFinalization(ctx context.Context, latestRound *types.DKGNetwork) error {
+	// Invalidate dealers that never submitted a deal during the dealing phase.
+	if k.isV190Round(ctx, latestRound) {
+		if err := k.invalidateMissingDealers(ctx, latestRound); err != nil {
```
stevemilk, line 18: Here we can only validate that deals exist but cannot promise the correctness because every deals will be recorded during `ProcessDeals`

0xHansLee: You're right. This is just for invalidating missed dealers. Even though the dealer submits invalid deal during dealing phase, it will be rejected and invalidated when finalization ends.

stevemilk (approved):

ramtinms (approved):

yingyangxu2026:
### Off-by-one found in a 4-node devnet regression: the highest-registration-index dealer is invalidated as "no deal submitted" every round

**Symptom**

Environment: devnet, 4 DKG validators (`total=4, threshold=2`), kernel `log-level=debug`.

`round 3` (initial bootstrap, `is_resharing=null`) — all 4 validators are `VERIFIED` and complete dealing normally:
- per-val kernel: `Succeed to generate deals round=3` + `Processed responses ... rejected=0`
- chain: `DKG dealing phase started round=3 total=4 threshold=2`, `Emitted BeginProcessDeals num_deals=12` (4 dealers x 3 recipients)
- val4 (registration `index=4`) story side: `Process deals complete round=3 submitted_deals=3 rejected_deals=0`

Yet at finalization:
```
WARN Invalidated dealer: no deal submitted during dealing phase round=3 validator=0xc025E30e... index=4
ERRO Failed to process DKGFinalized err="finalize DKG: dealer has been invalidated and cannot finalize"
```
`round 3` finalizes only `index {1,2,3}`; **`index 4` is wrongly invalidated**, shrinking the committee from 4 to 3. Per-round observation (`active` frozen at round 3):
```
r3 stage=4 reshare=-     total=4 thr=2 | finalized=[1,2,3] invalidated=[4]
r4 stage=5 reshare=true  total=4 thr=2 | finalized=[]      invalidated=[3,4]
r5 stage=5 reshare=true  total=4 thr=2 | finalized=[1]     invalidated=[2,3,4]
r6 stage=5 reshare=true  total=4 thr=2 | finalized=[2]     invalidated=[3,4]
r7 stage=5 reshare=true  total=4 thr=2 | finalized=[]      invalidated=[3,4]
r8 stage=3 reshare=true  total=4 thr=2 | finalized=[2]     invalidated=[1,3,4]
```
> Note: registration `index` is reassigned every round, so the same `index` across different rounds is not the same validator. `round 3`'s `invalidated=[4]` (highest index, with every validator dealing normally) is the cleanest signal of this bug. `round 4+` (resharing) invalidate additional indices each round and `active` never advances past 3 — that part mixes in resharing-specific failures and is out of scope for this issue.

**Root cause: 0-based vs 1-based dealer-index mismatch**

A deal's `Index` is **0-based all the way** from generation to `markDealersDealt`, but `invalidateMissingDealers` compares it against the **1-based** registration index, with no 0->1 conversion anywhere in between:

1. kyber `share/dkg/pedersen/dkg.go` `Deals()`: `&Deal{Index: uint32(d.oidx)}` — `d.oidx` is the dealer's **0-based** position in OldNodes.
2. story-kernel `service/helper.go` `createGenerateDealsResponse`: `pbDeal.Index = deal.Index` — passes the 0-based value through unchanged.
3. story `client/x/dkg/keeper/dkg_svc_dealing.go:111` `EnqueueDeals(resp.GetDeals())` + `queue.go` (`append`/`copy`) — **no conversion**.
4. `client/x/dkg/keeper/dkg_dealing.go` `markDealersDealt`:
   ```go
   if deal.Index < 1 || deal.Index > latestRound.Total { continue } // drops the 0-based index 0
   k.DealtDealers.Set(ctx, dealtDealerKey(latestRound.Round, deal.Index)) // stores the 0-based value
   ```
   The comment says `// deal.Index is the dealer's 1-based registration index`, but the value actually passed in is the **kyber 0-based index — that assumption is wrong**.
5. With n dealers the 0-based indices are `{0,1,…,n-1}`; after the `< 1` filter, `DealtDealers = {1,…,n-1}`.
6. `client/x/dkg/keeper/dkg_finalization.go` `invalidateMissingDealers` compares the 1-based `reg.Index in {1,…,n}`:
   - `index 1..n-1` happen to be present in `{1,…,n-1}` (a numeric coincidence — the ownership is actually shifted by one) -> pass;
   - `index n` is never present -> **invalidated**.

So **the dealer at the highest registration index (`n`) is wrongly flagged as no-deal every round** (`round 3`: `index=4`). The deal actually dropped by `< 1` is the 0-based `0` (= the dealer at registration index 1), but because the comparison is 1-based vs 0-based, the victim is shifted up to the highest index.

**Impact**

The highest-index dealer is invalidated every round. This is especially severe for small committees: a 3-node committee loses 1 -> only 2 finalized, and the next resharing round computes `CalculateThreshold(2)=1`, which kyber `vss.validT` (requires `t>=2`) rejects with `dealer: t 1 invalid` -> DKG cannot proceed.

**Suggested fix (make both sides agree on one convention)**
- **A (preferred)**: convert the dealer index to 1-based at the boundary — e.g. in kernel `createGenerateDealsResponse`, `Index: deal.Index + 1`, consistent with the existing 0<->1 convention already used for `RecipientIndex` / `session.Index`.
- **B**: keep 0-based on the chain side — drop the `< 1` skip in `markDealersDealt` (`0` is a valid dealer) and compare with `reg.Index - 1` in `invalidateMissingDealers`.

Worth adding an n>=4 bootstrap test that asserts all `n` dealers are recorded in `DealtDealers` (a test that builds deals with 1-based indices would hide this bug).

yingyangxu2026:
## participants root mismatch during DKG resharing

### Background

While testing "a 4-node committee losing 1 node → 3 nodes", DKG gets stuck: the **participants root** check fails at finalization.

**Environment**

| DKG address | IP | label |
| --- | --- | --- |
| `0xc8AE9Ca51859a4DB382EC0F4137A74155Cf936FB` | 20.222.19.88 | val1 |
| `0xC6cB20293dD838C3E839198B9F753c6Ed29e0c9B` | 172.207.250.203 | val2 |
| `0xa5D284381bB6905C6954802F1fc22C84E7d15Df6` | 104.41.162.2 | val3 (the stopped one) |
| `0xc025E30e4ff0A68cB623Ac671933B6FE99eF7Cc7` | 20.222.18.186 | val4 |

> In round 42, val1 was marked `Invalidated` during the dealing phase because of `no deal submitted` (the reason it failed to submit in time is still under investigation). This note focuses on the participants-root mismatch that this triggers.

### What is the participants root

At DKG finalization a commitment to the "set of effective participants for the round" is computed: take every effective participant (status `VERIFIED` or `FINALIZED`), sort the addresses, concatenate, and Keccak256 → a 32-byte root. The submitter (kernel) computes it and embeds it in the finalize tx; the chain recomputes it from the set **it** sees and requires the two to match, otherwise finalize is rejected. It is a consistency check against tampering with the participant set.

### Key logs

**① story side — finalize rejected, only 2 expected participants**

```text
2026-06-11T01:57:32.729 ERRO Failed to process DKGFinalized
  err="finalize DKG: failed to validate participants root: participants root mismatch"
  round=42
  expected_root=0x32322b0c706750861cef5dc42ac27d34f380ec01d23d1b6597b6f2b8b90090da
  got_root=0xdcc80c39164dc363a40146d18df19c9f7801f4bc9132e7e32ff88d4adba2ee4e
  num_participants=2
  stacktrace="[... dkg_handler.go:219 dkg_handler.go:134 dkg.go:168 dkg.go:30
               msg_server.go:170 ... baseapp.go:1021 ... abci.go:793 ... state.go:929 ...]"
```

- `num_participants=2` → the chain's expected set = `{v4, v2}` (2 members).
- The stacktrace shows this runs on the consensus path: cometbft `FinalizeBlock` → `ProcessDKGFinalized` → `validateParticipantsRoot`. So `expected_root` is the set agreed by consensus.

**② story side — val1 marked Invalidated at height 15261**

```text
01:57:27.895 ABCI call: FinalizeBlock              height=15261 proposer=9d28fd4
01:57:27.896 WARN Invalidated dealer: no deal submitted during dealing phase round=42 validator=0xc8AE9Ca5... index=2
01:57:27.896 Emitted BeginDKGFinalization event    round=42
01:57:27.897 Finalize call to kernel client        round=42
01:57:27.900 ABCI response: FinalizeBlock          val_updates=0 height=15261
```

The chain invalidates val1 at **height 15261 (01:57:27)**; from then on the on-chain effective set = `{v2, v4}`.

**③ kernel side — light-client trusted height lags (only reaches 15261 at 01:57:33; chain already at 15263)**

```text
D[2026-06-11|01:57:33.027] Verify non-adjacent newHeader against verifiedBlock
  trustedHeight=15261  newHeight=15263
```

When the kernel received the finalize request and computed the participants root (01:57:27), its trusted height had not yet reached 15261 (it only catches up to 15261 at 01:57:33), so it read the state from **before** val1 was invalidated. Note that the chain is already at `newHeight=15263` here — the kernel is **at least 2 blocks (~6s) behind**.

### Analysis — identical algorithm on both sides, only the input set differs

The chain's `validateParticipantsRoot` and the kernel's `calculateParticipantsRoot` use a byte-for-byte identical algorithm:

```text
take addresses with status VERIFIED + FINALIZED (exclude INVALIDATED)
→ lowercase & trim → sort → concat 20 bytes each → Keccak256
```

Same algorithm, different root → the input participant set must differ.

### Computed — the difference is val1 (not the stopped val3)

Enumerating combinations of the 4 addresses with that algorithm matches both roots exactly:

| root | value | set |
| --- | --- | --- |
| `got_root` (kernel, embedded in finalize tx) | `0xdcc80c39...` | `{v4, v2, v1}` — 3 |
| `expected_root` (chain, consensus) | `0x32322b0c...` | `{v4, v2}` — 2 |

**The kernel still counts val1 as an effective participant; the chain does not.**

### Conclusion

The mismatch happens because **the kernel and the chain read chain state at different heights, and val1's Invalidate state-change falls exactly between those two heights**:

- chain (`expected`): validates at a height **after** val1 was removed → `{v2, v4}`, `num_participants=2`;
- kernel (`got`): computes the root at its lagging trusted height (**before** val1 was removed) → `{v1, v2, v4}`;
- the two snapshots are separated by "val1 invalidated at 15261" → the sets differ by val1 → root mismatch → finalize rejected.

yingyangxu2026:
## `invalidateMissingDealers` uses the wrong index space in resharing rounds: it treats the OLD committee's dealer index as the NEW round's registration index

### What this is

The `markDealersDealt` + `invalidateMissingDealers` mechanism added in this PR implicitly assumes `deal.Index + 1 == reg.Index` *within the same round*. That holds for the **initial DKG** (a single committee where dealers and members are the same set, so the index spaces are isomorphic). It does **not** hold for **resharing rounds**.

The kernel itself documents this (`service/dkg_process_deals.go:93-95`):

```go
// dealerCount upper-bounds Deal.Index. Non-resharing has a single
// committee acting as dealers; resharing dealers are members of the
// PRIOR active committee (kyber's c.OldNodes inside the next-DKG).
...
dealerCount = int(latest.GetTotal())   // resharing branch = OLD committee total
```

So in a resharing round, `deal.Index` lives in the **OLD committee's** index space, while `invalidateMissingDealers` (`dkg_finalization.go:210`) walks the **NEW round's** registrations and compares against the **NEW** `reg.Index`. The two index spaces are re-assigned by registration arrival order between rounds (`getNextDKGRegistrationIndex = len(regs)+1`, `dkg_registration.go:70`), so they are permuted differently — the comparison is meaningless.

### Observed (reproduced on devnet): node removal 4→3 wrongly invalidates a healthy validator

Environment: devnet, 4 DKG validators, running the fixed #838 (`7350827c`). Round 40 healthy (`total=4`). We stopped one validator's kernel (`val3`), triggering a 4→3 resharing.

On-chain registrations for the two rounds (`/dkg/registrations`):

```
round 40 (OLD committee = the dealers)        round 42 (NEW committee)
  reg.index=1  val1  FIN  (kyber 0)             reg.index=1  val2  VERIFIED
  reg.index=2  val3  FIN  (kyber 1)  STOPPED    reg.index=2  val1  INVALIDATED  <-- wrongly killed
  reg.index=3  val4  FIN  (kyber 2)             reg.index=3  val4  FINALIZED
  reg.index=4  val2  FIN  (kyber 3)
```

(addresses: val1=`0xc8AE…`, val2=`0xC6cB…`, val3=`0xa5D2…` stopped, val4=`0xc025…`)

The dealers that actually submitted deals are val1/val4/val2 (val3 is stopped), with OLD kyber indices `0 / 2 / 3`.

`markDealersDealt` runs against the **NEW** round (`latestRound.Total = 3`), so the bounds check uses the new total:

| dealer | OLD kyber idx | `idx >= 3` ? | key written |
| --- | --- | --- | --- |
| val1 | 0 | no | **1** |
| val4 | 2 | no | **3** |
| val2 | 3 | **yes → dropped** (second defect: new total used to bound an OLD index) | — |
| val3 | 1 | stopped, no deal | — |

→ `DealtDealers[42] = {1, 3}`

`invalidateMissingDealers` then walks round-42 by NEW index: val2(1) ✅ kept, **val1(2) ❌ → INVALIDATED**, val4(3) ✅ kept.

**Why val1 specifically:** the stopped node is val3 (OLD kyber idx 1 → it would have marked key 2). Nobody marks key 2. In the NEW committee, key 2 == new index 2 == **val1**. The hole left by the stopped dealer falls on whoever happens to sit at that NEW index — which was val1. val1 takes the bullet for val3 purely because of the old↔new index permutation.

Downstream: val1 is flipped to INVALIDATED *during the finalization window*, so the kernel (reading a lagging light-client height where val1 is still VERIFIED) and the chain (val1 already removed) compute different participant sets → **participants root mismatch** → finalize is rejected for several rounds (rounds 41–44 all show `gpk=null`), self-healing only at round 45 (total down to 3). This is the upstream cause of the participants-root-mismatch failure reported earlier.

### Code-level prediction (NOT yet reproduced): node addition / expansion fails structurally

The dealt-key set is always a subset of `{1 .. oldTotal}` (bounded by the number of OLD dealers). In an expansion, the NEW committee has `newTotal > oldTotal` registrations, and a newly-joining node has no old share so it **correctly submits no deal**. Therefore every registration whose NEW index falls in `(oldTotal, newTotal]` can never match a dealt key and would be INVALIDATED. Predicted consequences:

- the committee can never grow past `oldTotal` — a single expansion round cannot complete;
- each added node kills whichever validator registered **last** in that round (`index == newTotal`) — decided by a registration-tx arrival race, so it is **not necessarily the joiner**; it can decapitate an existing member;
- subsequent resharing rounds retry into the same wall → stuck-round loop.

I have not reproduced the expansion case yet; the above is derived from the code paths cited.

Issue #850, opened 2026-06-24
DKG resharing: a validator absent last round is still expected to deal after it rejoins → always INVALIDATED (re-joiner-as-dealer)


yingyangxu2026:
## Summary

In DKG resharing, a validator that was **absent in the previous (prev-active) round** — i.e. it did not participate and therefore holds no sealed share for that round — is still **expected to deal** when it comes back online and registers for the next round. But a resharing dealer must reshare using its **previous round's sealed key share**, which a returning node does not have, so its `GenerateDeals` fails, it submits no deal, and `invalidateMissingDealers` marks it INVALIDATED.

Root cause: **`expectedDealers` uses the staking `ActiveValSet` (all bonded & un-jailed validators), not the set of validators that actually hold the previous round's DKG share.** A still-bonded re-joiner that was absent last round is wrongly expected to deal.

> ⚠️ This round's INVALIDATED also **triggers a downstream per-round index-misattribution cascade** (see the sibling issue), so it is not just "one extra node invalidated" — it is the entry point of the cascade.

## Environment

- **story**: `regression5` integration branch (`dkg/dev` + #837/#838/#839/#840/#847/#848, commit `964edc4`).
- **story-kernel**: `debug/dkg-readdition-trace` (= `main` + #78/#80 + trace logs; CC_B MRENCLAVE `170e218d`).
- Topology: 4 TEE validators + 1 non-TEE; DKG `total=4 / threshold=2`.
- Periods: `reg=50 / deal=80 / fin=80 / active=200` (blocks); devnet 2.5s/block.

## Steps to reproduce

1. Bring up a 4-node DKG chain and let it run stable resharing (here rounds 3–18 all FINALIZED).
2. **Stop one low-voting-power TEE validator's (val3) `story-kernel` service** (keep its `story`/`node-geth` so it stays bonded and un-jailed — it is merely an absent DKG dealer). The chain shrinks cleanly to `total=3` (rounds 19/20, val3 ABSENT), GPK unchanged.
3. **Restart val3's kernel** after `active=19` so it re-registers in round 21.

## Observed

`dkg-rounds-status.sh` (active=28, with RegIndex):

```
round | val1(0xc8AE9C)           | val2(0xC6cB20)           | val3(0xa5D284)           | val4(0xc025E3)           | GPK16
[...]

PR #852, opened 2026-06-26
fix(dkg): derive resharing expected dealers from prev round share holders

0xHansLee:
## Problem

In a resharing round `expectedDealers` used the previous active round's staking `ActiveValSet` as the dealer set. A validator that stayed bonded but was absent from the previous DKG round holds no share and cannot reshare, yet it was expected to deal and got `INVALIDATED` when it rejoined and registered for the next round.

## Fix

Derive the resharing expected dealers from the previous round's participant registrations (`VERIFIED`/`FINALIZED`) — the actual share holders — so an absent re-joiner that holds no previous share is not expected to deal. This matches the kernel's prev-round dealer committee.

## Tests

- `TestInvalidateMissingDealers_AbsentRejoinerNotExpected` — a bonded validator absent from the previous round, rejoining the next round, is not invalidated.

issue: #850

### DIFF
--- a/client/x/dkg/keeper/dkg_finalization.go
+++ b/client/x/dkg/keeper/dkg_finalization.go
@@ -284,7 +284,22 @@ func (k *Keeper) expectedDealers(ctx context.Context, latestRound *types.DKGNetw
 			return nil, nil
 		}
 
-		return prevActive.ActiveValSet, nil
+		// Resharing dealers are the previous round's FINALIZED registrations, not the staking
+		// active set: only a finalized member holds a valid share and can reshare, so an absent
+		// re-joiner (or a non-finalized member) is not expected to deal.
+		prevRegs, err := k.getDKGRegistrationsByRound(ctx, prevActive.Round)
+		if err != nil {
+			return nil, errors.Wrap(err, "failed to fetch previous round registrations")
+		}
+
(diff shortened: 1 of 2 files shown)


yingyangxu2026 (approved):

client/x/dkg/keeper/dkg_finalization.go
```diff
@@ -284,7 +284,23 @@ func (k *Keeper) expectedDealers(ctx context.Context, latestRound *types.DKGNetw
...
+		// and cannot reshare.
+		prevRegs, err := k.getDKGRegistrationsByRound(ctx, prevActive.Round)
+		if err != nil {
+			return nil, errors.Wrap(err, "failed to fetch previous round registrations")
+		}
+
+		dealers := make([]string, 0, len(prevRegs))
+		for i := range prevRegs {
+			if prevRegs[i].Status == types.DKGRegStatusVerified ||
+				prevRegs[i].Status == types.DKGRegStatusFinalized {
```
yingyangxu2026: Only validators with the status `types.DKGRegStatusFinalized` in the last active round are allowed to send a valid deal, so I believe we do not need to add `types.DKGRegStatusVerified` to the `dealers` list.

Issue #854, opened 2026-07-16
TDH2 decrypt worker is never restarted when a node restarts after the next DKG round opened registration — Aeneid decryption fully down since the geth v1.2.2 upgrade restart


yingyangxu2026:
## Summary

The TDH2 decrypt worker is only (re)started when the **latest** DKG round is in the ACTIVE stage. If a node restarts while the *next* round has already opened registration (latest round != active round), no code path ever starts the worker again: decrypt requests keep queueing forever and the validator submits **zero partial decryptions** until the next round activates — which on Aeneid's cadence is days away.

This took CDR decryption on Aeneid fully down on 2026-07-16: the geth v1.2.2 security-patch upgrade restarted `story` on all 5 DKG validators at ~01:07 UTC, while round 30 had been in REGISTRATION since ~19:30 UTC the previous day (active round = 29). Every `accessCDR` has failed since.

## Environment

- Aeneid, the 5 DKG validators (jpe-aeneid-validator1..5)
- story `v1.7.0-stable` (commit `6a97606`) under cosmovisor; kernel 0.1.0

## Symptom

- cdr-sdk integration run 29482061608 (2026-07-16 08:03 UTC): every test depending on partial decryptions times out (single-wallet `accessCDR` end-to-end ~136s timeouts, `queryCDRPartials`/`decryptDataKey` ~76s, 100-ephemeral-wallet shared-vault test 100/100 failed), while all write-path tests pass (uploader 12/12, observer 18/18) and read txs land on-chain.
- On-chain: round 29 ACTIVE (total=5, threshold=3, GPK present), round 30 in REGISTRATION with 5/5 registrations — so kernels, attestation and the registration path are all healthy. Only partial-decryption submissions stopped.

## Node-side evidence (validator5)

- `story` restarted at **01:07:58 UTC** (upgrade window), under `cosmovisor.service`.
- The kernel is healthy: light client verifying current heights, `Connected to kernel endpoint` at 01:08:01, no light-client errors after 01:10.
- Since the restart the only decrypt-related lines are the queueing ones, with the pending counter climbing monotonically and nothing ever draining it:

```
26-07-16 08:42:43 INFO Handling threshold decrypt request  round=29 label=...25c9 block_height=20958209
26-07-16 08:42:43 INFO Queued threshold decrypt request    session=29 round=29 label=...25c9 pending_requests=211
```

[...]

PR #855, opened 2026-07-20
fix(dkg): restart decrypt worker whenever an active round exists

0xHansLee:
## Problem

The TDH2 decrypt worker was only (re)started for the **latest** round. If a node restarted while the next round had already opened registration (latest round != active round), no start path fired: `ResumeDKGService`'s recovery branch gates on the latest round's stage being `Active`, which it isn't once the next round opened. Decrypt requests then queued indefinitely until the next round activated — days on Aeneid. This took Aeneid CDR decryption fully down after the geth `v1.2.2` upgrade restart.

## Fix

In `ResumeDKGService`, (re)start the idempotent worker whenever an **active** DKG round exists, independent of the latest round's stage. `LatestActiveRound` is written only on the finalize success path, so it stays intact across the next round's churn and across restarts.

## Tests

- `TestResumeDKGService_ActiveRoundStartsWorker`
- `TestResumeDKGService_NoActiveRoundNoWorker`
- `TestResumeDKGService_CompletedActiveStartsWorker`

issue: #854

### DIFF
--- a/client/x/dkg/keeper/dkg_svc.go
+++ b/client/x/dkg/keeper/dkg_svc.go
@@ -100,6 +100,14 @@ var decryptWorkerRunning atomic.Bool
 
 // ResumeDKGService reloads unfinished DKG sessions and resumes their execution safely without spawning duplicate goroutines.
 func (k *Keeper) ResumeDKGService(ctx context.Context, dkgNetwork *types.DKGNetwork) {
+	// The decrypt worker serves the latest ACTIVE round, which may differ from the
+	// latest round once the next round has opened. StartDecryptWorker is idempotent.
+	if active, err := k.getLatestActiveDKGNetwork(ctx); err != nil {
+		log.Warn(ctx, "Failed to get latest active DKG round while resuming decrypt worker", err)
+	} else if active != nil {
+		k.StartDecryptWorker()
+	}
+
 	session, err := k.stateManager.GetSession(dkgNetwork.Round)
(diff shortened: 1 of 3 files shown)


yingyangxu2026:
Verified the fix on a devnet: restarting `story` in the window where the next round has opened registration while an active round still serves (latest != active) now brings the decrypt worker back up ("Decrypt worker started" + the worker picks up the active round's queue), which the old stage-gated path skipped. One remaining gap in the same failure family:

The worker has no deadline anywhere downstream. `workerCtx` is `context.Background()` (`dkg_svc.go:295`), passed straight into `client.PartialDecryptTDH2` (`:632`), and `kernel_client.go:59` dials with only transport creds — no per-call timeout, no keepalive. If the kernel wedges (accepts the connection but never responds), the worker's synchronous loop blocks indefinitely and never panics, so the `decryptWorkerRunning` flag is never reset and `ResumeDKGService`'s CompareAndSwap keeps failing — the worker stays dead with no self-heal, the same end state as #854. Suggest a per-call timeout context plus gRPC keepalive dial params (and/or a watchdog that resets the flag on a stalled tick).

Also, when there's no active round (worker not started, `:107`) or the CompareAndSwap no-ops because the worker is already running (`:289-291`), nothing is logged, so a healthy-but-idle worker can't be told apart from a stuck flag over a dead goroutine. A Debug line at each would help (Info would be too noisy given it runs every block).

PR #857, opened 2026-07-21
fix(dkg): do not complete an active-round session without key material

0xHansLee:
## Problem
`resumeFailedSession` force-completed any failed active-round session, even one with empty `GlobalPubKey`/`PubKeyShare`. The node marked itself finished while contributing zero partial decryptions.

## Fix
Complete only sessions that hold key material. A keyless active-round session is recovered node-locally: recompute and seal the share via the kernel (`callTEEFinalizeDKG`, **no** on-chain finalize vote for an already-active round). The recovery is guarded by:
- the per-round DKG service lock (`tryAcquireDKGSvc`), preventing concurrent recovery goroutines from racing on the shared session; and
- a check that the recovered `GlobalPubKey` matches the on-chain consensus key before completing.

On failure the session is left failed and logged. Node-local only — no consensus/`KVStore` state is touched.

## Tests
- `TestResumeFailedSession_ActiveStage_RecoversKeyMaterial`, `_MissingKeyMaterial`, `_DivergentKey`
- `TestRecoverActiveSessionKeyMaterial_ConcurrentSingleFinalize` (`-race`)

issue: none

### DIFF
--- a/client/x/dkg/keeper/dkg_svc.go
+++ b/client/x/dkg/keeper/dkg_svc.go
@@ -98,6 +98,10 @@ const dkgAsyncTimeout = 2 * time.Minute
 // contract-call timeout in contract_client.go.
 const decryptKernelCallTimeout = 60 * time.Second
 
+// maxActiveRecoveryAttempts caps keyless active-round recovery attempts; past it the session
+// is left failed and an escalation alert is logged instead of retrying forever. Node-local.
+const maxActiveRecoveryAttempts = 5
+
 // dkgAsyncContext creates a new context for async DKG service goroutines with a timeout.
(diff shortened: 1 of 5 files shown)


yingyangxu2026:
Every failure path calls `MarkFailed`, which leaves the session `PhaseFailed`, so the next block's `ResumeDKGService` re-enters recovery and runs `callTEEFinalizeDKG` again — the block-level retry has no backoff and no attempt cap (the kernel call has internal retries, but nothing between blocks). A persistently-failing kernel, or a genuinely divergent key, turns into a per-block finalize retry indefinitely, visible only via the "does not match" error log. Worth adding a backoff plus an escalation/alert after N attempts.

PR #858, opened 2026-07-21
fix(dkg): prune dkg sessions older than the retention window

0xHansLee:
## Problem
Local DKG sessions (the in-memory map and on-disk `session_*.json` files) were never pruned and grew unbounded. The existing `CleanupExpiredSessions` was dead code and would have deleted the active completed session.

## Fix
Add `PruneOldSessions`, wired into round activation, deleting sessions for rounds older than `activeRound - sessionRetentionRounds` (`= 10`). Node-local only (no `KVStore`/consensus access); the round-based criterion is deterministic with no wall-clock dependence. Retention of 10 rounds comfortably exceeds the decrypt path's 2-round lookback.

## Tests
- `TestStateManager_PruneOldSessions`, `_RetainedRoundStillUsable`, `_NoopBelowHorizon`

issue: none

### DIFF
--- a/client/x/dkg/keeper/dkg_svc_complete.go
+++ b/client/x/dkg/keeper/dkg_svc_complete.go
@@ -29,6 +29,10 @@ func (k *Keeper) handleDKGComplete(ctx context.Context, dkgNetwork *types.DKGNet
 		return
 	}
 
+	// A round just activated, so bound session growth by pruning rounds well below it.
+	// session.Round is the latest active round and is always retained.
+	k.stateManager.PruneOldSessions(ctx, session.Round)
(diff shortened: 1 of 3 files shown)


yingyangxu2026:
Two things. The commit message says the window keeps "the previous `sessionRetentionRounds` (3) rounds," but the constant is actually `10` — the message needs updating. Separately (pre-existing, not introduced here): a crash-orphaned `session_<round>.json.tmp` is never cleaned up — `loadSessions` globs `session_*.json` and the prune's `os.Remove` targets the non-tmp name, so those files linger indefinitely.

